PR #1, opened 2020-03-24
Old Project Code Migration

finben:


Hymns/ContentView.swift
```diff
@@ -9,11 +9,25 @@
...
+                       NavigationView {
+                            VStack {
+                                SearchBar(text: $searchText)
+                                List {
+                                    ForEach(hymns.filter { self.searchText.isEmpty ? true : $0.contains(self.searchText)}, id: \.self) { hymn in
+                                        Text(hymn)
+                                }
+                                }.navigationBarTitle(Text("Look up any hymn"))
+                            }
+                        }
```
skywalkerdude: nit: indenting looks funny. could you fix?

finben: Done, and I learned a valuable lesson here. When you make file changes like I did adding two files then you have to submit the xcodeproj file to github. Otherwise xcode does not pick the files up when you pull them into your local. So always submit xcodeproj when adding or deleting files.

Hymns/SearchBarView.swift
```diff
@@ -0,0 +1,42 @@
+//
+//  SearchBarView.swift
+//  Hymns
+//
+//  Created by Johanna Findeisen on 3/25/20.
+//  Copyright © 2020 skywalkerdude. All rights reserved.
+//
+
```
skywalkerdude: Do we generally want to have these header files? In the past I haven't included them

finben: Good point, yeah let's do away with them.

PR #2, opened 2020-03-26
Tab views

finben:
So I set up the UI tab view structure and added TODO view files for the other views we will need. One thing to note is I tried to use the FIGMA icons but neither the PNG or SVG files work with the tab bar. I can put them in as images but they do not light up when selected since they do not conform to the SF Symbols standards defined by Apple. To get our custom icons working right we need to have them meet certain requirements. I will mark this as an issue for later but for now we can use default icons until we are able to get our custom icons in.

Please make sure this works on your local before merging.

### DIFF
--- a/Hymns/ContentView.swift
+++ b/Hymns/ContentView.swift
@@ -16,7 +16,7 @@ struct ContentView: View {
 
             FavoritesView().tabItem {
                 Image(systemName: "heart")
-            }.tag(2)
+                }.tag(2)
 
             SettingsView().tabItem {
                 Image(systemName: "gear")
--- a/Hymns/ContentView.swift
+++ b/Hymns/ContentView.swift
@@ -16,7 +16,7 @@ struct ContentView: View {
 
             FavoritesView().tabItem {
                 Image(systemName: "heart")
-                }.tag(2)
+            }.tag(2)
 
             SettingsView().tabItem {
                 Image(systemName: "gear")
--- a/Hymns/ContentView.swift
+++ b/Hymns/ContentView.swift
@@ -1,32 +1,29 @@
-//
-//  ContentView.swift
-//  Hymns
-//
-//  Created by Luke Lu on 3/23/20.
-//  Copyright © 2020 skywalkerdude. All rights reserved.
-//
-
 import SwiftUI
 
 struct ContentView: View {
-    var hymns = hymnTestData
-    @State private var searchText : String = ""
-
+    @State var selected = 0
 
     var body: some View {
-        NavigationView {
-            VStack {
-                SearchBar(text: $searchText)
-                List {
-                    ForEach(hymns.filter { self.searchText.isEmpty ? true : $0.contains(self.searchText)}, id: \.self) { hymn in
-                                        Text(hymn)
-                    }
-                }.navigationBarTitle(Text("Look up any hymn"))
-            }
+        TabView(selection: $selected) {
+            //systemName: basically means they are coming from SF Icons.
+            SearchView().tabItem {
+                Image(systemName: "magnifyingglass")
+            }.tag(0)
+
+            BrowseView().tabItem {
+                Image(systemName: "book")
+            }.tag(1)
+
+            FavoritesView().tabItem {
+                Image(systemName: "heart")
+            }.tag(2)
+
+            SettingsView().tabItem {
+                Image(systemName: "gear")
+            }.tag(3)
         }
     }
 }
-
 
 struct ContentView_Previews: PreviewProvider {
     static var previews: some View {
--- a/Hymns/FavoritesView.swift
+++ b/Hymns/FavoritesView.swift
@@ -2,7 +2,11 @@ import SwiftUI
 
 struct FavoritesView: View {
     var body: some View {
-        Text("TODO: Favorites")
+        NavigationView {
+            VStack {
+                Text("TODO: Favorites")
+            }.navigationBarTitle("Favorites")
+        }
     }
 }
 
--- a/Hymns/FavoritesView.swift
+++ b/Hymns/FavoritesView.swift
@@ -1,0 +1,13 @@
+import SwiftUI
+
+struct FavoritesView: View {
+    var body: some View {
+        Text("TODO: Favorites")
+    }
+}
+
+struct FavoritesView_Previews: PreviewProvider {
+    static var previews: some View {
+        FavoritesView()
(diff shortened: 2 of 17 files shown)


Hymns/FavoritesView.swift
```diff
@@ -0,0 +1,17 @@
...
+struct FavoritesView: View {
+    var body: some View {
+        NavigationView {
+            VStack {
+                Text("TODO: Favorites")
+            }.navigationBarTitle("Favorites")
+        }
+    }
+}
+
```
skywalkerdude: Do you need the

#if DEBUG
...
#endif

I saw it in the video

(comment applies to all the files)

Hymns/ContentView.swift
```diff
@@ -1,33 +1,30 @@
...
+                Image(systemName: "gear")
+            }.tag(3)
         }
     }
 }
 
-
 struct ContentView_Previews: PreviewProvider {
     static var previews: some View {
         ContentView()
```
skywalkerdude: Could you wrap this in a Group{} so we can see more in the preview?

```
Group { 
    ContentView(selected: 0)
    ContentView(selected: 1)
    ContentView(selected: 2)
    ContentView(selected: 3)
}
```

Hymns/SceneDelegate.swift
```diff
@@ -66,6 +57,4 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
         (UIApplication.shared.delegate as? AppDelegate)?.saveContext()
     }
 
-
```
skywalkerdude: Thanks for fixing the formatting!

Hymns/SearchView.swift
```diff
@@ -0,0 +1,30 @@
+import SwiftUI
+
+struct SearchView: View {
+    var hymns = hymnTestData
+    @State private var searchText: String = ""
+
+    //Currently we have to use UIKIT to change the unselected tabicons from grey to black.
+    init() {
+        UITabBar.appearance().unselectedItemTintColor = UIColor.black
```
skywalkerdude: I am a bit confused why UITabBar's appearance lives in SearchView. Shouldn't it be in ContentView? That seems to be the container of all the tabs right, while this is just the search view?

Hymns/SearchView.swift
```diff
@@ -0,0 +1,30 @@
+import SwiftUI
+
+struct SearchView: View {
```
skywalkerdude: I would rename this to HomeView (and the container to HomeContainer) or something like that. When you click the search bar, it should take you to a new view to actually do the search, and that one makes more sense to be the SearchView. Just bringing it up now so we don't have to refactor later

Hymns/SearchView.swift
```diff
@@ -0,0 +1,30 @@
...
+    
+    var body: some View {
+        NavigationView {
+            VStack {
+                SearchBar(text: $searchText)
+                List {
+                    ForEach(hymns.filter { self.searchText.isEmpty ? true : $0.contains(self.searchText)}, id: \.self) { hymn in
+                        Text(hymn)
+                    }
+                }.navigationBarTitle(Text("Look up any hymn"))
```
skywalkerdude: If it's not too much trouble, could we put these in some kind of localized file? Instead of hardcoding the strings?

Hymns/ContentView.swift
```diff
@@ -1,33 +1,30 @@
...
-                List {
-                    ForEach(hymns.filter { self.searchText.isEmpty ? true : $0.contains(self.searchText)}, id: \.self) { hymn in
-                                        Text(hymn)
-                    }
-                }.navigationBarTitle(Text("Look up any hymn"))
-            }
+        TabView(selection: $selected) {
+            //systemName: basically means they are coming from SF Icons.
+            SearchView().tabItem {
+                Image(systemName: "magnifyingglass")
```
skywalkerdude: I think these SF icons look fine. If these are the recommended iOS icons, then we should just use that to maintain a consistent feel across the platform. No need to import custom icons if there are provided ones :)

Hymns/FavoritesView.swift
finben: Yes, from what I have been able to figure out thus far we can use them. The prievews are by default set to this in the more recent versions of swiftUI that we are using. So they automatically only run when in debug. I am not sure how to apply the #if DEBUG in the best way here though. Maybe we can get this pushed in and I'll look into this in the near future?

Hymns/ContentView.swift
finben: Nice! I didn't know about that. I addressed in commit 7a95b7b

Hymns/SearchView.swift
finben: Yeah, good question. This has been driving me absolutely crazy. I have spent hours trying to figure out why this will not work on our Home Screen view... I'm still not sure how to get it working on the main screen view at this point. I went ahead and moved it to HomeView which makes at least a little more sense then being in searchview.

Hymns/SearchView.swift
finben: Done, good idea!

Hymns/SearchView.swift
finben: I went ahead and add the localized files. So the way they work is quite nice. We still leave this hardcoded but it will automatically convert to the right text based on what we define in our localized string files.

Hymns/ContentView.swift
finben: Sweet! I agree!

Hymns/ContentView.swift
```diff
@@ -1,33 +1,30 @@
...
-//  ContentView.swift
-//  Hymns
-//
-//  Created by Luke Lu on 3/23/20.
-//  Copyright © 2020 skywalkerdude. All rights reserved.
-//
-
 import SwiftUI
 
 struct ContentView: View {
```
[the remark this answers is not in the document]
finben: Done, sounds good!

finben:
To address your first question, I will start to look into unit testing more. So far apparently Apple has not released anything to help unit test with SwiftUI because of how new it is. One guy developed a framework to try and do unit testing https://github.com/nalexn/ViewInspector . I also read some people don't see it as necessary for reasons I don't fully understand yet. I'll use the coming days to try and look into this and understand what we should do for testing more on this project. I've never done much with unit/ui testing so it is all a bit unfamiliar for me still at this point.

PR #8, opened 2020-04-02
Hook up initial data layer to populate the HymnLyricsView with live data from the server

skywalkerdude:
1) Create networking component to fetch songs from HymnalNetApi
2) Create repository of hymns that will handle fetching from the api
3) Create a view model to display only the lyrics of a hymn
4) Hook up everything up via a dependency injection framework (Resolver https://github.com/hmlongco/Resolver)


Hymns/Display/HymnLyricsView.swift
```diff
@@ -0,0 +1,29 @@
+import SwiftUI
+import Resolver
+
+public struct HymnLyricsView: View {
+    
+    @ObservedObject private var viewModel: HymnLyricsViewModel
+    
+    init(viewModel: HymnLyricsViewModel) {
```
finben, line 8: What is the benefit of using init(.....)     instead of just having viewModel: HymnLyricsViewModel?

Hymns/Display/HymnLyricsView.swift
```diff
@@ -0,0 +1,29 @@
...
+    public var body: some View {
+        guard let lyrics = viewModel.lyrics else {
+            return Text("error!")
+        }
+        
+        guard !lyrics.isEmpty else {
+            return Text("loading...")
+        }
+
+        return Text(lyrics[0].verseContent[0])
```
finben, line 21: How does this work? I can see that verseContent would need to be an array but why does lyrics have to be an array?

Hymns/Display/HymnLyricsViewModel.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation
+import Resolver
+
+class HymnLyricsViewModel: ObservableObject {
+    
+    @Published var lyrics: [Verse]? = [Verse]()
+    
+    @Injected private var hymnsRepository: HymnsRepository
```
finben, line 8: Can you explain @Injected? I get that it is from Resolve?

Hymns/Display/HymnLyricsViewModel.swift
```diff
@@ -0,0 +1,26 @@
...
+                self.lyrics = nil
+                return
+            }
+            
+            self.lyrics = hymn.lyrics
+        }
+    }
+}
+
+extension Resolver {
```
finben, line 22: How does this work?

Hymns/Display/HymnLyricsView.swift
skywalkerdude: A song has a list of verse and each verse is a list of strings (lines). Lyrics = list of verses

Hymns/Display/HymnLyricsView.swift
skywalkerdude: I'm not quite sure what you are asking.

If you are asking why we can't just do private var viewModel: HymnLyricsViewModel = HymnLyricsViewMode(), that's because we are injecting the view model via dependency injection

If you are asking why we can't just do @injected var viewModel: HymnLyricsViewModel, it's because we need to create the HymnLyricsView from the container view, so (afaik, and I might be wrong), we can use the magical @injected incantation because we are manually creating that view, and thus have to pass the dependencies through the constructor

Hymns/Display/HymnLyricsViewModel.swift
skywalkerdude: Admittedly this is a bit of a magical invocation. Basically what it does under the hood is that it creates getters/setters, and the DI framework will set the variable behind the scenes, so you will just have it.

I'm unsure if we want to go this route or if we want to fully go the constructor injection route. It is less code in terms of boilerplate and set-up, but we'll see if it gives us problems in the future.

Hymns/Display/HymnLyricsViewModel.swift
skywalkerdude: Here is how extensions work: https://www.hackingwithswift.com/sixty/9/3/extensions

We are extending the Resolver class to have extra methods that essentially "registers" our dependencies, so they can be available via the resolve() method and the @Injected parameter.

PR #13, opened 2020-04-03
Just to get the stanza to fully display properly. 

finben:
Small pull request just to get the stanzas displaying


Hymns/Display/HymnLyricsView.swift
```diff
@@ -11,20 +11,28 @@ public struct HymnLyricsView: View {
...
+        
+        return AnyView(
+            ForEach(lyrics, id: \.self) { verse in
+                VStack(alignment: .leading) {
+                    Spacer()
+                    ForEach(verse.verseContent, id: \.self) { line in Text(line)
+                    }
+                }
+            }
+        )
```
skywalkerdude, line 29: This looked a little funky when I built it, especially in the Previews. For some reach each verse was its own screen.

Also, if you put the spacer on top, that means short songs will all get smushed to the bottom. Would something like this work:

```suggestion
        
        return AnyView(
            VStack(alignment: .leading) {
                ForEach(lyrics, id: \.self) { verse in
                    Group {
                        ForEach(verse.verseContent, id: \.self) { line in
                            Text(line)
                        }
                    }
                }
            }
        )
```

Then after doing that, all you need to do is fix the formatting to look like the figma spec

finben: Nice! I added a further commit expanding off of what you did here. I think the spacing is starting to look good. Also, I tested with changing the accessibility to large system font etc. and everything seems to come out looking pretty good.  

One question about the figma spec is that Apple really seems to be promoting dynamic type which is in SwiftUI by default that changes their font size with the system. If we lock in a font size and spacing to a specific font size like 16pt for example then the lyrics will no longer adjust with the system. I was thinking we could still have options like your other app but the options would only change dynamic fonts from like normal body to large etc... Idk what you think?

finben:
Previews are great! I spend some time looking at getting previews running last night too lol... All I could get was the loading screen up. Good job!

finben:
Because of Dynamic Type in IOS you may want to consider if supporting different text sizings is something still worth doing? The text should automatically be sizing to whatever their preference text size is locally? However, it may still be worth it if they are sharing the screen or something with someone sitting next to them?

PR #20, opened 2020-04-05
First attempt at solving hymn stanza numbers. Also, at getting margin…

finben:
* Stuck at the dilima of needing to loop through stanzas and put in a stanza number but not being able to change state within a ForEach since it is a view that is dependent on state not changing. Not yet working but the margins are formatting is getting closer. If you have any idea how to solve the hymn stanza numbering problem let me know.


Hymns/Display/HymnLyricsView.swift
```diff
@@ -21,18 +28,23 @@ public struct HymnLyricsView: View {
...
-                    Group {
-                        ForEach(lyrics, id: \.self) { verse in
+                    ForEach(lyrics, id: \.self) { verse in
+                        VStack { //needed to contain verticle spacer
                             Group {
-                                ForEach(verse.verseContent, id: \.self) { line in
-                                    Text(line)
-                                }
-                                Spacer().frame(height: 30)
+                                HStack(alignment: .top) { //needed for verse numbers to be displayed to the left
```
skywalkerdude, line 34: Instead of creating an HStack to contain the verse number, is it possible just to create a separate view for line number and just align it with the verse content?

finben: Not easy, I tried a solution like that. My main problem is in the actual getting line numbers based off the hymn and how to put in a blank when the stanza is a chorus. I am wondering if there is a solution where we dynamically create an array based on the stanza types. For example if the chorus is the second stanza it would look something like [1, " ", 2, 3, 4]. Then during the ForEach put them in. But I still haven't found a viable solution. It's very difficult because state can not change during the ForEach loop so you have to bring you data ready to be pasted in instead of creating during the loop.

skywalkerdude: I just pushed a VerseView.swift onto this branch. Do you think something like that could work?

Hymns/Display/VerseView.swift
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    var body: some View {
+        VStack(alignment: .leading) {
+            ForEach(0..<verseLines.count) { index in
+                if self.verseNumber != nil && index == 0 {
+                    VerseLineView(verseNumber: self.verseNumber!, verseText: self.verseLines[index])
+                } else {
+                    VerseLineView(verseText: self.verseLines[index])
+                }
```
skywalkerdude, line 52: Admittedly this if-else block looks more like java than swift code. If you could make this more ternay/swift-like, that'd be helpful. I've found that using the `if let` syntax doesn't really work in ForEach blocks

PR #24, opened 2020-04-07
Create navigation from the home screen to the lyrics screen

skywalkerdude:
Create navigation from the home screen to the lyrics screen. I deviated a little bit from ray wenderlich's article (https://www.raywenderlich.com/4161005-mvvm-with-combine-tutorial-for-ios) in the sense that I just pass the view-to-navigate-to through the SongResult.

If I did it the way that the article did it, we would have the SongResult creating the HymnsLyricsView. That doesn't make much sense because the SongResult is literally just a song result, and it shouldn't have to know about the HymnLyricsView. This way, the song result is much more generic and just passes along the navigate-to information without having to worry about what it actually is. Rather, the thing that creates the song result worries about that, which makes sense because that view model is creating the component, so it should know where it should go.

Haha, hope that made sense. Let me know if you have any questions!


Hymns/Common/SongResultViewModel.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+import SwiftUI
+
+class SongResultViewModel<DestinationView>: ObservableObject, Identifiable where DestinationView: View {
```
finben, line 4: I see it is an ObservableObject but what takes advantage of that? I don't see any @published variables or any places that has SongResultViewModel as an ObservedObject? Will this come into play later?

Hymns/Home/HomeView.swift
```diff
@@ -2,29 +2,40 @@ import Resolver
...
 
-                //This list does two things. It filters all the Hymns based on search function and it also provides the linking to their respective detailed views.
                 List {
-                    ForEach(self.allHymns.filter { self.searchText.isEmpty ? true : $0.songTitle.contains(self.searchText)}) { hymn in
-                        NavigationLink(destination: HymnLyricsView(viewModel: Resolver.resolve())) {
-                            Text(hymn.songTitle)}
+                    ForEach(self.viewModel.recentSongs) { recentSong in
+                        NavigationLink(destination: recentSong.destinationView) {
+                            SongResultView(viewModel: recentSong)
+                        }
```
finben, line 24: So let me see if I understand this, by calling recentSong.destinationView we are going through our HomeViewModel published variable recentSong to then use the SongResultViewModel which then by virtue of being of a generic type that conforms to a view. It is able to be set as a destination and brings us to the HymnLyricsView for that given hymn.... My head is spinning trying to understand this thing. Where does the SongResultView come into play?

Hymns/Common/SongResultViewModel.swift
skywalkerdude: you're right. It doesn't need to be an ObservableObject

Hymns/Home/HomeView.swift
skywalkerdude: Yeah, you're right.

So here's HomeViewModel:
`class HomeViewModel: ObservableObject {

    @Published var recentSongs: [SongResultViewModel<HymnLyricsView>] = [SongResultViewModel<HymnLyricsView>]()

    init(recentSongs: [SongResultViewModel<HymnLyricsView>]) {
        // TODO fetch recent songs instead of passing it in
        self.recentSongs = recentSongs
    }
}`

Here, we're passing in the SongResultViewModel because we don't have any data source yet. But in the future, we'd be making a call to the database to get the song results, and we'd create the SongResultViewModel based on the song result model from the db.

An example would be:
`    init() {
        var dbModels: RecentSongDatabseModel = database.getRecentSongs()
        recentSongs = dbModels.map(SongResultViewModel)
    }`

So you get the db models, and create the data model here, which contains the navigation information

skywalkerdude: Feel free to ping me on messenger if you want me to explain it more. Happy to hop on a call or explain over chat if that wasn't clear

PR #25, opened 2020-04-07
Updating all of our localization

finben:
So this updates all of the localization that I could find so far that is hardcoded into our app. This supports German and Spanish. My suggestion ongoing though is that whenever we add any string into our app we just also add it to the localization file for English. Then at the end of our development we find someone fluent in all of the languages we want to support to go through and translate for the other languages. So going forward whenever any words are hardcoded into the app we just simultaneously update the localization en file. This will be an easy way to maintain keeping up with our localization as we work. 

In the end we really need someone who is fluent to go through by hand and make sure all of our strings translate in the absolute best way.


finben:
Yeah, exactly. As far as I know unless we do some grep type find and
replace we would have to update each one by hand if we change something
later. I think it's probably one of the last things we should do on the
project besides just storing the strings in the English localization file
as we work. Then later on we can populate all of the other languages we
want to support and have the text translated.

skywalkerdude:
Interesting. I thought we could just do something like this: https://swiftwithmajid.com/2019/10/16/localization-in-swiftui/ and use the localized keys isntead?

But yeah we can wait till later to localize everything. I've just localized it as I went on Android, but we don't necessarily have to do that for iOS

finben:
When you look at documentation for Text you can see that it takes
LocalizedStringKey not a String into its initializer. So it actually is
localizable by default. That is why even without putting that localized
string key. I think it was an update with swift or something like that. So
it was already doing that functionality at least on everything that is
within a Text ().

By doing the English file we are essentially setting up the keys for our
whole app. So we can do that as we go. then for the other languages we just
copy paste the entire English localization file and put in the new
translations for that language. And SwiftUI will automatically handle
everything else.

finben:
Okay, found out the best workflow for this I think. There is a really cool way built in to xcode to automatically export all of the localized keys. Check it out 

To localise your app you need:
1. Use SwiftUI elements like: Text("Nice"). Nice becomes the key in Localizable.strings.  (Note: This is by default because as I mentioned above Text automatically conforms to the protocol)
2. In the cases where text is not in the SwiftUI element needs to use NSLocalizedString.
3. Export localisation: Edit -> Export for Localisation...
4. Give exported files to translator.
5. Import translations: Edit -> Import Localisations...

PR #29, opened 2020-04-09
Fixes our current GitHub Action. Also, adds a TestPlan

finben:
Okay, a few things to note. Basically, the swift.yaml now works but I wasn't sure if I should delete the old build and test? I'm not sure if they are still necessary or not but didn't want to delete unless I was sure. But the new ones work. I also specified them to work with a specific test plan. If you don't know what test plans are yet then checkout https://developer.apple.com/videos/play/wwdc2019/413/ starting at time stamp 27:15. Having a test plan will let us check tests on and off easily which may be nice to have as our test suite grows. Also, lets say that we don't want ui testing to always run on our CI then we can easily check it off in the test plan and our GitHub action will update right away. 

Because having a test plan means you have to pair it with a scheme I went ahead and checked in the xcshareddata otherwise I don't think it would work right out of the box for you. So give it a try. Hopefully it works well.

### DIFF
--- a/.github/workflows/swift.yml
+++ b/.github/workflows/swift.yml
@@ -25,3 +25,11 @@ jobs:
       run: swift build -v
     - name: Run tests
       run: swift test -v
+      
+# CI Testing that actually works!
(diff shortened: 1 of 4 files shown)


finben:
Lol... Idk I think it must be an artifact. I don't have anything like that on my local either.

PR #35, opened 2020-04-10
Makes searchbar first responder when you go into search view.

finben:
This makes the view autoselect the searchbar like in android app when you go to searchview


Hymns/Common/SearchBar.swift
```diff
@@ -45,6 +47,10 @@ struct SearchBar: UIViewRepresentable {
 
     func updateUIView(_ uiView: UISearchBar, context: UIViewRepresentableContext<SearchBar>) {
         uiView.text = text
+        if isFirstResponder && !context.coordinator.didBecomeFirstResponder {
```
skywalkerdude, line 50: What does first responder do?

skywalkerdude: Instead of passing this in, could we just make an extension of it in the search view? That way we don't need to pass this in from the home screen

finben: First responder is just a bool that we can set, whenever it is true then the search bar will be the first thing selected in a view. 

We still don't need to pass it in since it has a default initialization value of false. It is optional to pass it in as an argument. That is why I passed it in from searchview but not from home screen. Is that what you mean?

skywalkerdude: ahh, I see. okay, sounds good then! I didn't see the default value :)

Hymns/Search/SearchView.swift
```diff
@@ -8,7 +8,7 @@ struct SearchView: View {
     var body: some View {
         NavigationView {
             VStack {
-                SearchBar(text: $viewModel.searchInput)
+                SearchBar(text: $viewModel.searchInput, isFirstResponder: true)
```
skywalkerdude, line 11: Do you think we can make this a more informative name? like "showKeyboard" or something, since that's essentially what we're doing. isFirstResponder is a little bit vague

skywalkerdude: other than that lgtm!!!!

PR #40, opened 2020-04-11
Incorporate Quick and Nimble into our project

skywalkerdude:
Add Quick and Nimble into the project for better unit tests. See https://www.raywenderlich.com/135-behavior-driven-testing-tutorial-for-ios-with-quick-nimble for a good tutorial on what they are and how to use them

For some reason, a simple pod install doesn't work, so we are forced to add a pod deintegrate into the build step. There is an issue reported in the Nimble library and I also commented on the issue (https://github.com/Quick/Nimble/issues/702#issuecomment-612366941). Until we get more guidance on how to fix this CwlCatchException.h not found error, we're stuck deintegrating and re-installing all pods on every build. :(


finben:
Lol, so are we going back to cocoapods?

ikesyo:
The reason for the problem is https://github.com/Quick/Nimble/tree/8.x-branch/Carthage/Checkouts/CwlPreconditionTesting/Dependencies is ignored due to https://github.com/HymnalDreamTeam/Hymns-iOS/blob/f1363ad71ee2a9ce47dcbe34d3b372762fb6dbef/.gitignore#L69-L70. I recommend you change the entry to `/Dependencies`.

PR #45, opened 2020-04-12
Navigate with hymn lyrics inside detail hymn screen instead of the other way around

finben:
This addresses the comment you made on my other PR about storing hymn lyrics within detail hymn screen as the container instead of what we have currently. The only thing is right now the navigation naming is pretty confusing since HymnLyricsViewModel will now be observed by detailed hymn screen and then the view Hymn lyrics view is called. 

Let me know what you think how we can reorg the names and files to make it a little more straightforward. But functionality works.


Hymns/DetailHymnScreen.swift
```diff
@@ -34,14 +35,33 @@ struct DetailHymnScreen: View {
...
 
 struct DetailHymnScreen_Previews: PreviewProvider {
 
     static var previews: some View {
-        DetailHymnScreen()
+        //DetailHymnScreen()
+
+        let classic1151 = HymnLyricsViewModel(hymnToDisplay: PreviewHymnIdentifiers.hymn1151, hymnsRepository: Resolver.resolve(), mainQueue: Resolver.resolve(name: "main"))
+        classic1151.lyrics = classic1151_preview.lyrics
+        let classic1151View = HymnLyricsView(viewModel: classic1151)
```
skywalkerdude, line 56: I think these need to be DetailHymnScreens

Hymns/DetailHymnScreen.swift
```diff
@@ -3,6 +3,7 @@ import Resolver
 
 struct DetailHymnScreen: View {
     @State var favorited: Bool = false
+    @ObservedObject private var viewModel: HymnLyricsViewModel // Pass through to Hymn Lyrics View
```
skywalkerdude, line 6: I think instead of observing the HymnLyricsViewModel, it'd make more to observe a DetailHymnScreenViewModel and have that expose the HymnLyricsViewModel for the HymnLyricsView to consume.

In the future, DetailHymnScreenViewModel would expose potentially 4 view models:
HymnLyricsViewModel
HymnChordsViewModel
HymnGuitarViewModel
HymnPianoViewModel

Hymns/Display/HymnLyricsView.swift
```diff
@@ -16,7 +17,7 @@ public struct HymnLyricsView: View {
 
         return AnyView(
             VStack {
-                DetailHymnScreen()
+               // DetailHymnScreen()
```
skywalkerdude, line 20: delete?

Hymns/DetailHymnScreen.swift
```diff
@@ -34,14 +35,33 @@ struct DetailHymnScreen: View {
...
+    init(viewModel: HymnLyricsViewModel) {
+        self.viewModel = viewModel
+    }
 }
 
 struct DetailHymnScreen_Previews: PreviewProvider {
 
     static var previews: some View {
-        DetailHymnScreen()
+        //DetailHymnScreen()
```
skywalkerdude, line 52: delete?

Hymns/Display/HymnLyricsView.swift
```diff
@@ -3,7 +3,8 @@ import Resolver
 
 public struct HymnLyricsView: View {
 
-    @ObservedObject private var viewModel: HymnLyricsViewModel
+    //Pass in from DetailHymnScreen
```
skywalkerdude, line 6: I'm unclear why this is no longer observed?  Is it because there's no binding on the lyrics?

If that's the case, can we also remove the @Published on the view model as well

finben: Yeah, I don't think it needs to be observed at this point since there is nothing binding and it should never change within the view. If there is any change to this later we can make it observed again. And yeah, I think we can also remove the @published.

Hymns/DetailHymnScreen.swift
finben: yep

Hymns/Display/HymnLyricsView.swift
finben: yeah lol

Hymns/DetailHymnScreen.swift
finben: Okay, I took a stab at restructuring this. Lmk what you think. I added a detailedHymnScreenVM to expose HymnLyricsViewModel

Hymns/DetailHymnScreen.swift
finben: addressed in recent commit

Hymns/Display/HymnLyricsView.swift
skywalkerdude, line 6: just FYI. In general, I don't think we need comments like this, since it effectively couples the hymn lyrics view with the detail hymn view which is what we are trying to avoid with the View Model pattern. But it's okay, it doesn't need to block this PR. We can just fix it as we continue working

PR #51, opened 2020-04-14
enable swift lint again

skywalkerdude:


Dangerfile.swift
```diff
@@ -3,7 +3,8 @@ import Danger
 
 let danger = Danger()
 
-// FAILS with https://github.com/danger/swift/issues/339 SwiftLint.lint(configFile: ".swiftlint.yml", swiftlintPath: "Pods/SwiftLint/swiftlint")
+let file: File = "Hymns/Display/HymnLyricsView.swift"
```
f-meloni, line 6: It looks correct, but just to be sure I would use `danger.git.modifiedFiles.filter { $0.hasPrefix("Hymns") }`
That should already have the correct path in it

skywalkerdude: I did as you said, but I still get the same error :/

f-meloni: I wonder if is because the compiled version of swiftlint in the repo doesn't work on a Linux machine. 
Would be interesting build it in the container with Danger and see what happens

I have this repo https://github.com/f-meloni/danger-swift-swiftlint-action with a Dockerfile that installs both Danger and Swiftlint, you can make a try with that, and see what happens

skywalkerdude: I copied what's in the README into my danger.yml and copied over the contents of the Dockerfile, but there seems to be an error since none of the check are running. Did I do it correctly? 

https://github.com/HymnalDreamTeam/Hymns-iOS/pull/51/commits/7193aaacb44a207311b2598a8e8979ae6aa12af8

f-meloni: So you don't need the Dockerfile, and you need to remove the `swiftlintPath: "Pods/SwiftLint/swiftlint"` in order to make it get the installed version of swiftlint

f-meloni: The only thing is that is not triggering the workflow at all, at least from what I see here, am I wrong? 🤔

f-meloni: I've updated that to use the latest danger-swift and swiftlint, try to update the version to 0.2.6

skywalkerdude: Yeah. I guess when I rebased up to the latest master, it started triggering again.
One weird thing to note:
This commit passed: https://github.com/HymnalDreamTeam/Hymns-iOS/pull/51/commits/6be78cb6a10056e2231a2b69ca8ce67f60662fd3
But when I added a random comma space to trigger a swiftlint warning, it fails:
https://github.com/HymnalDreamTeam/Hymns-iOS/pull/51/commits/69bac0b003c4350d174cca44a8ceadb6e23efaa5

skywalkerdude: Added your action: https://github.com/HymnalDreamTeam/Hymns-iOS/pull/51/commits/e90511fc6a92dbb5c6d7a4e478005b4823accebb
It looks like that check isn't being triggered

skywalkerdude: Added your action: https://github.com/HymnalDreamTeam/Hymns-iOS/pull/51/commits/7bc05d6a6385d40cd3a99bf851f87770e756bca1

and it seems to be working! Do recommend I just use this instead of the the regular dagger-swift action? Does this run all the danger or just swiftlint?

f-meloni: This action builds danger-swift from master and, separately, swiftlint, then you can use it as you normally use danger-swift.
I’m going to release an already built version of it on dockerhub, that should not build it every time, and save a lot of CI time.

PR #52, opened 2020-04-14
Setting up the detail toolbar for tags

finben:
Addresses multiple things

1. Fix's a navigation bug that we had because of having two navigationViews causing nesting. Needed to delete the navigation view that was in settings.

2. Moved our root navigation view to the home container in order to make it the parent of the TabView instead of the other way around. This allows us to finally not show the tab view on every single view thereafter. 

3. Set up a detail toolbar that sits on the detailed hymns screen. I will also be trying to create custom sheets to pop up on click for adding things like tags.


finben:
Okay, that's fine. Are you going to have a media player in the alpha launch? Do we still need any of the bottom toolbar or just let it be for now? That sounds good. I'll try to do the favorites. 

There are still a few things in this PR worth implementing though like fixing the navigation problem, and a few other small things. I will throw them up on another PR after the one you have right now goes in.

skywalkerdude:
There probably won't be a media player or bottom toolbar. It'll be pretty bare bones, just to get the most common critical user journey up and running for emily to use and whoever else wants to use it

finben:
Okay, I think that is a good idea. It will be good just to have it out in production.

PR #53, opened 2020-04-14
Some UI work related to displaying a recent hymn from the home view

skywalkerdude:
1. Rename DetailHymnScreen to DisplayHymnView (which isn't necessarily a better name, but it's just more consistent with the Android naming)
2. Make it a modal to be displayed from the recents screen (so it's full screen)
3. Remove generics and use type erasure instead to handle generics and protocols with associated types in Swift (https://www.swiftbysundell.com/articles/type-erasure-using-closures-in-swift/)


Hymns/Infrastructure/View+Extensions.swift
```diff
@@ -0,0 +1,12 @@
+import Foundation
+import SwiftUI
+
+extension View {
+
+    /**
+     * Use type erasure as a way to return a generic protocol as a type, instead of forcing a concrete type (https://www.swiftbysundell.com/articles/type-erasure-using-closures-in-swift/).
```
finben, line 7: Nice, I see you found SwiftBySundell

Hymns/Display/DisplayHymnView.swift
```diff
@@ -1,8 +1,9 @@
 import SwiftUI
 import Resolver
 
-struct DetailHymnScreen: View {
-    @ObservedObject private var viewModel: DetailHymnScreenViewModel
+struct DisplayHymnView: View {
```
finben, line 4: I think this still needs to be observed. Once @ObservedObject is removed we no longer have access to changing the @published property of a hymn being favorited or not. Is there a particular reason for removing?

skywalkerdude: ah, okay. I didn't think about the favorites. Okay. let's keep it. Do you want to do it? or should I?

finben: I already did in my most recent commit. I just tagged a commit on to your PR. Mainly to fix a nested nav issue and a few other small superficial things. Take a look. I approve your stuff.

PR #55, opened 2020-04-15
Utilize default parameters to inject dependencies by default, so we don't need to pass them in anymore. Makes code so much cleaner. Huzzah!

skywalkerdude:
Utilize default parameters to inject dependencies by default, so we don't need to pass them in anymore. Makes code so much cleaner. Huzzah!


skywalkerdude:
sweet! Thanks!

PR #56, opened 2020-04-16
Fix up navigation on the home container and create proper back navigation on the display hymns screen.

skywalkerdude:
Fix up navigation on the home container and create proper back navigation on the display hymns screen. Getting closer to an MVP! Also there's finally enough functionality to write some UI tests


Hymns/Display/DisplayHymnView.swift
```diff
@@ -3,47 +3,35 @@ import Resolver
...
-                    HStack {
-                        NavigationLink(destination: HomeContainerView()) {
-                            Image(systemName: "xmark").accentColor(Color.black)
-                        }
-                        Spacer()
-                        if self.viewModel.favorited == true {
-                        }
-                        //TODO: Favorited needs to actually do something when clicked other than just say "favorited"
-                        Spacer()
-                        Button(action: {self.viewModel.toggleFavorited()}) {
```
finben, line 21: Any reason in particular you nixed my heart button from working? I can put it back in later but just wanted to check

finben, line 21: Any reason in particular you nixed my heart button from working? I can put it back in later but just wanted to check

Hymns/Display/DisplayHymnView.swift
```diff
@@ -3,47 +3,35 @@ import Resolver
...
-        }.navigationBarTitle("", displayMode: .inline).navigationBarHidden(true)
-    }
-
-    init(viewModel: DisplayHymnViewModel) {
-        self.viewModel = viewModel
+            Spacer()
+        }
+        .hideNavigationBar()
+        .onAppear {
+            self.viewModel.fetchHymn()
```
finben, line 33: This is cool. So it does the fetch on appear. Nice

Hymns/Display/DisplayHymnView.swift
skywalkerdude: ah oops. Might have been an oversight on my part. Could you add it back later, when you add the favorites with realm?

skywalkerdude: ah oops. Might have been an oversight on my part. Could you add it back later, when you add the favorites with realm?

Hymns/Display/DisplayHymnView.swift
skywalkerdude: Yeah, haha. That way it won't be making like hundreds of requests to my free heroku server and overloading it 😂

PR #60, opened 2020-04-17
Make the settings view a little more generic by creating a base class and performing type erasure so we can display a list of generic settings

skywalkerdude:
Make the settings view a little more generic by creating a base class and performing type erasure so we can display a list of generic settings


finben:
Take a look at my commit when you get the chance. I struggled to get it to work with your generic type stuff lol... Anyways I want you to see a draft first. We can go sheet route or just do a destination link. But to me it forsure looks nicer in webkit and it keeps the user from leaving our app and just seems more flawless of a user experience.

skywalkerdude:
Nice! I incorporated your changes!

PR #62, opened 2020-04-19
Basically re-implement TabView so we can have an animated tab bar indicator on the bottom of each tab

skywalkerdude:
Basically re-implement TabView so we can have an animated tab bar indicator on the bottom of each tab

### DIFF
--- a/Hymns/Common/TabView/TabItem.swift
+++ b/Hymns/Common/TabView/TabItem.swift
@@ -1,0 +1,12 @@
+import Foundation
+import SwiftUI
+
+public protocol TabItem: Identifiable, Equatable {
+
+    associatedtype Label: View
+    associatedtype Content: View
+
+    var label: Label { get }
+    var a11yLabel: Text { get }
+    var content: Content { get }
+}
--- a/Hymns/Common/TabView/TabItem.swift
+++ b/Hymns/Common/TabView/TabItem.swift
@@ -3,10 +3,12 @@ import SwiftUI
 
 public protocol TabItem: Identifiable, Equatable {
 
-    associatedtype Label: View
+    associatedtype SelectedLabel: View
+    associatedtype UnselectedLabel: View
     associatedtype Content: View
 
-    var label: Label { get }
+    var selectedLabel: SelectedLabel { get }
+    var unselectedLabel: UnselectedLabel { get }
     var a11yLabel: Text { get }
     var content: Content { get }
 }
--- a/Hymns/Home/HomeContainerView.swift
+++ b/Hymns/Home/HomeContainerView.swift
@@ -12,8 +12,7 @@ struct HomeContainerView: View {
                 .browse,
                 .favorites,
                 .settings
-                ]).hideNavigationBar()
-            .edgesIgnoringSafeArea(.top)
+            ]).hideNavigationBar()
         }
     }
 }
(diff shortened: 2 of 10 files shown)


Hymns/Common/TabView/TabItem.swift
```diff
@@ -0,0 +1,14 @@
...
+
+public protocol TabItem: Identifiable, Equatable {
+
+    associatedtype SelectedLabel: View
+    associatedtype UnselectedLabel: View
+    associatedtype Content: View
+
+    var selectedLabel: SelectedLabel { get }
+    var unselectedLabel: UnselectedLabel { get }
+    var a11yLabel: Text { get }
```
finben, line 12: Strange naming? ally label instead of a11lyLabel

Hymns/Home/HomeContainerView.swift
```diff
@@ -7,49 +7,38 @@ struct HomeContainerView: View {
...
-                // Make the unselected tabs black insetad of grey.
-                UITabBar.appearance().unselectedItemTintColor = .black
-            }.edgesIgnoringSafeArea(.top)
+            IndicatorTabView(currentTab: $selectedTab, tabItems: [
+                .home,
+                .browse,
+                .favorites,
+                .settings
+                ]).hideNavigationBar()
+            .edgesIgnoringSafeArea(.top)
```
finben, line 16: I don't think we should ignore safe area top. To me it looks better with the space.

Hymns/Settings/SettingsView.swift
```diff
@@ -10,13 +10,12 @@ struct SettingsView: View {
 
     var body: some View {
         VStack(alignment: .leading) {
-            CustomTitle(title: "Settings")
+            Text("Settings").customTitleLayout().padding(.top)
```
finben, line 13: Not sure, what this is for. The problem is it throws off consistency with browse and favorites

Hymns/Home/HomeView.swift
```diff
@@ -1,17 +1,18 @@
...
     @ObservedObject private var viewModel: HomeViewModel
 
-    init(viewModel: HomeViewModel) {
+    init(viewModel: HomeViewModel = Resolver.resolve()) {
         self.viewModel = viewModel
     }
 
     var body: some View {
         VStack(alignment: .leading, spacing: 10) {
-            Text("Look up any hymn").customTitleLayout()
```
finben, line 14: Probably should also make this a CustomTitle to match with browse, favorites, and settings. It looks to me like on the android app they were all the same text sizing and even placement. The one that is hard to tell is this one the Look up any hymn. Is it the same font size and location. Hard to tell.

Hymns/Common/TabView/TabItem.swift
skywalkerdude: it's not ally, it's a11y. Are you saying a11y is weird?

Hymns/Home/HomeView.swift
skywalkerdude: I'm pretty sure they are the same font and size and placement. But I fix according to your comments to make it even more obvious/explicit. let me know if these address your comments

Hymns/Common/TabView/TabItem.swift
finben: yeah, just seemed like a strange naming to me. But if thats what you want no problem

skywalkerdude: a11y is shorthand for accessibility, so since this label is used for accessibility, it's the accessibilityLabel, or a11yLabel for short

https://whatisa11y.org/

PR #64, opened 2020-04-20
Reimplement the tab bar plurely in swift ui and add animations that will hide the title and show the cancel button when the search bar is clicked

skywalkerdude:
Reimplement the tab bar plurely in swift ui and add animations that will hide the title and show the cancel button when the search bar is clicked. Further work will be to remove the SearchView/ViiewModel and consolidate in HomeView/ViewModel instead.

### DIFF
--- a/Hymns/Common/SearchBar.swift
+++ b/Hymns/Common/SearchBar.swift
@@ -1,67 +1,78 @@
 import SwiftUI
 
-struct SearchBar: UIViewRepresentable {
+/**
+ * Custom search bar that will animate in a `Cancel` button when activated/selected.
+ * https://stackoverflow.com/questions/56490963/how-to-display-a-search-bar-with-swiftui
+ */
+struct SearchBar: View {
 
-    @Binding var text: String
-    var selectedOnAppear: Bool = false //when true will make a search bar first responder
+    @Binding var searchText: String
+    @Binding var searchActive: Bool
+    let placeholderText: String
 
-    class Coordinator: NSObject, UISearchBarDelegate {
+    var body: some View {
+        HStack {
+            HStack {
+                Image(systemName: "magnifyingglass")
+                TextField(placeholderText, text: $searchText, onEditingChanged: { _ in
+                    withAnimation(.easeInOut(duration: 0.3)) {
+                        self.searchActive = true
+                    }
+                }, onCommit: {
+                    print("onCommit")
+                }).foregroundColor(.primary)
 
-        @Binding var text: String
-        var didBecomeFirstResponder = false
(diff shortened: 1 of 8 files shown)


Hymns/Common/SearchBar.swift
```diff
@@ -1,67 +1,78 @@
...
+                    // this must be placed before the other commands here
+                    UIApplication.shared.endEditing(true)
+                    self.searchText = ""
+                    withAnimation {
+                        self.searchActive = false
+                    }
+                }
+                .foregroundColor(Color(.systemBlue))
+                .transition(AnyTransition.move(edge: .trailing))
+                .animation(.easeInOut(duration: 0.3))
```
finben, line 43: .3 feels a bit slow especially considering a user is ready to quickly look up a hymn. Looks great though as far as the animation. I tested .1 and .2 and to me it seems like .2 is an optimal amount of animation time. Up to you though, just something to consider

Hymns/Home/HomeView.swift
```diff
@@ -3,32 +3,35 @@ import SwiftUI
...
+                CustomTitle(title: "Look up any hymn")
+            }
+
+            SearchBar(
+                searchText: $searchText,
+                searchActive: $searchActive,
+                placeholderText: "Search by numbers or words")
+                .padding(.horizontal)
+
+            (searchActive ? Text("Recent searches") : Text("Recent hymns")).font(.caption).padding(.top).padding(.leading)
```
finben, line 26: Another thing that may just be my personal view. But to me swapping the text here just feels distracting to the experience. All of a sudden you see the text change and then you wonder what did that text say before? And so you go back to try and see what just changed. I would just leave them both to the more generic "Recent". Just less distracting and I don't think anyone would ever question what it means by recent.

skywalkerdude:
Yeah, because I commented out the search code for now because I still have to migrate it. BUt in an effort to make the PR shorter and more manageable, I split it out into two tasks. We can re-introduce the search functionality in a follow-up

finben:
Pretty dope overall! And nice that it's moving towards a more SwiftUI approach

Hymns/Home/HomeView.swift
skywalkerdude: Yeah I thought it was a bit janky too. but Emily suggested making the text more specific to recent hymns and recent searches, since they are technically different things. I'm inclined to just keep it for now, until we get something she can look at and see her suggestions. It's easy to remove/change, so we don't lose out on anything if we keep it while if we change it now, we might forget about discussing it later

finben: Sure, thats fine. Yeah, I think what makes it so smooth is that it really does not feel like a view change when you go into search mode. But with the text change you kind of get the feeling like the view just changed more. Not as smooth. We can talk to Emily about it later.

PR #68, opened 2020-04-21
Favorites

finben:
Migration of old PR to avoid Git Merge issues I was having


skywalkerdude:
looks like the status checks are broken?

finben:
Alot of stuff is broken right now :( For some reason now whenever I go into a favorited hymn from favorites it immediately pops right back to the list....

finben:
GitHub was also down this morning with issues too though. So this might be GitHub being broken. They are having issues of their own this morning.

PR #70, opened 2020-04-21
For whatever reason it is the custom tabbar that kills favorites

finben:
Well, this is a joke somewhat in the fact I don't really mean to merge this. However, it does fix the problem lol. Something about the custom tab bar just does not work with favorites and is causing the odd behavior. I have no idea why this is... may be a dumb SwiftUI navigation thing. Super annoying.


skywalkerdude:
shall we close this request, @finben ?

finben:
Yeah, thanks!

PR #71, opened 2020-04-23
Realm2

skywalkerdude:
tried to incorporate both favorites and recents into relam... not without bugs though


styusuf:
Looks like none of our pull requests are able to pass that check :(

finben:
My most recent commit fixes a bug where if you favorite a hymn while the hymn is still loading it stores an empty hymn as a favorite. The commit makes the favorite toggle only available after fetch hymn has completed and gathered a title. It, also looks a bit better imo because the favorite button comes in with the hymn instead of being there beforehand

Hymns/Display/DisplayHymnView.swift
```diff
@@ -22,19 +22,23 @@ struct DisplayHymnView: View {
                 Text(viewModel.title)
                     .fontWeight(.semibold)
                 Spacer()
+                if viewModel.title.isEmpty {
+                    EmptyView()
+                } else {
                 Button(action: {self.viewModel.toggleFavorited()
```
skywalkerdude, line 28: ohh great find! I was wondering why i got an empty favorite sometimes

Could we put this logic in the View Model instead? We can put the fetchFavoriteStatus call inside of the receiveValue, then we won't fetch the favorites until after the hymn come back

Hymns/Display/DisplayHymnView.swift
```diff
@@ -22,19 +22,23 @@ struct DisplayHymnView: View {
...
             }
             Spacer()
             HymnLyricsView(viewModel: self.viewModel.hymnLyricsViewModel)
             Spacer()
         }
         .hideNavigationBar()
         .onAppear {
-            self.viewModel.fetchFavoriteStatus()
             self.viewModel.fetchHymn()
+            self.viewModel.fetchFavoriteStatus()
```
skywalkerdude, line 41: then we can remove this line too

Hymns/Display/DisplayHymnView.swift
finben: So, I tried a few things. But as soon as I stopped using the viewModel.title there was always a lack of sync between when the title would come in and the heart toggle. It is split seconds but still noticeable. As long as I leave it viewModel.title it is smooth and they come in at the exact same time the title and the favorites heart. Idk, you can take a stab at it if you want but this was my experience.

PR #73, opened 2020-04-24
activity indicator on loading

finben:
Basically from the resource you linked me on the trello board. Pretty simple but smooth.


Hymns/Display/HymnLyricsView.swift
```diff
@@ -3,6 +3,7 @@ import SwiftUI
 public struct HymnLyricsView: View {
 
     @ObservedObject private var viewModel: HymnLyricsViewModel
+    @State private var shouldAnimate = true
```
skywalkerdude, line 6: Do we need this state variable if we never toggle it to false?

What happens if you just get rid of the state variable and make the ActivityIndicator always be animating. Then whenever lyrics.isEmpty is false, it'll just re-render the view without the activity indicator and we never have to keep track of the indicator load state

finben: Yeah, I thought about that. That's fine with me. I left it how I saw it in case we needed to use this loader somewhere else but for now I'll change it to always be on.

skywalkerdude: if this is always true, then we can delete this variable

PR #79, opened 2020-04-24
Try to make Danger work with plugins

skywalkerdude:


Dangerfile.swift
```diff
@@ -20,3 +21,5 @@ var bigPRThreshold = 1000;
 if (danger.github.pullRequest.additions! + danger.github.pullRequest.deletions! > bigPRThreshold) {
     warn("> Pull Request size seems relatively large. If this Pull Request contains multiple changes, please split each into separate PR will helps faster, easier review.");
 }
+
+Coverage.xcodeBuildCoverage(.derivedDataFolder("Build"), minimumCoverage: 50, excludedTargets: ["DangerSwiftCoverageTests.xctest"])
```
f-meloni, line 25: I'm not sure you are generating any coverage here, in order to get the coverage you need to build first, also I don't think you have `DangerSwiftCoverageTests.xtest` in your target

.github/workflows/danger.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Danger
-        uses: danger/swift@3.0.0
+        uses: f-meloni/danger-swift-swiftlint-action@0.2.8
```
f-meloni, line 13: If you use `docker://frmeloni/danger-swift-with-swiftlint:1.1.0` you should save more than 6 minutes of build

Dangerfile.swift
skywalkerdude: Yeah, so I'm unsure how to generate the coverage data. When I try to add an xcodebuild test command to the danger.yml, it complains that xcodebuild is not found. I suppose it's because it's on ubuntu and not macos? But when I change it to run on macos, it complains that danger only runs on ubuntu.

I do have a test scheme (HymnsTests), but I don't see an .xctest anywhere, so I thought maybe it was just danger-swift thing.

skywalkerdude: @f-meloni do you know how I can generate the coverage data?

f-meloni: If you want to use xcode you need to be on mac os, this doesn't work docker.
You can generate the data and then pass it to the instance with danger https://help.github.com/en/actions/configuring-and-managing-workflows/persisting-workflow-data-using-artifacts#passing-data-between-jobs-in-a-workflow

Or you can use the SPM configuration https://github.com/danger/swift#swift-package-manager-more-performant

skywalkerdude: @f-meloni I did what you suggested and passed the data from the mac-os test run(https://github.com/HymnalDreamTeam/Hymns-iOS/pull/79/checks?check_run_id=665825697), but I'm still getting the same error I've always been getting:

�[0;0m�[31mERROR: Dangerfile eval failed at Dangerfile.swift
�[0;0m�[31mERROR: Could not get the results JSON file at /tmp/danger-response.json

I've found that if I just comment out the line that uses the plugin, then it passes no problem. Could it be something wrong with the way that DangerSwift uses Marathon to import the plugins?

skywalkerdude: @f-melonii If you look at https://github.com/HymnalDreamTeam/Hymns-iOS/pull/139, it's the same error but instead of coverage, I'm trying to get the xcode build summary

PR #81, opened 2020-04-25
another attempt at pagination. Loads better and wraps. But has an biz…

finben:


finben:
This is the approch from Apple but the problem is it is used when adding SwiftUI into UIKit. I haven't been able to reverse engineer it since our project is rooted in SwiftUI. It displays much faster than the other approach I took though. It works fine when it is the rootView. But as soon as you try to navigate into it, I get a bizarre index error. I think it is because it is trying to put SwiftUI into UIKit instead of the other way around.

https://developer.apple.com/tutorials/swiftui/interfacing-with-uikit

PR #84, opened 2020-04-26
Add activity indicator to search screen also

skywalkerdude:
Add activity indicator to search screen also

### DIFF
--- a/Hymns/Home/HomeView.swift
+++ b/Hymns/Home/HomeView.swift
@@ -30,7 +30,7 @@ struct HomeView: View {
                     Spacer()
                     HStack {
                         Spacer()
-                        ActivityIndicator().eraseToAnyView()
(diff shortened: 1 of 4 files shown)


Hymns/Home/HomeView.swift
```diff
@@ -25,11 +25,23 @@ struct HomeView: View {
...
-                    SongResultView(viewModel: songResult)
+            if self.viewModel.isLoading {
+                VStack {
+                    Spacer()
+                    HStack {
+                        Spacer()
+                        ActivityIndicator().eraseToAnyView()
+                        Spacer()
+                    }
+                    Spacer()
```
skywalkerdude, line 36: If you have a better way to center the activity indicator, let me know

finben: That should be fine.

PR #85, opened 2020-04-26
add Firebase Craslytics crash reporting

skywalkerdude:
add Firebase Craslytics crash reporting


Hymns.xcodeproj/xcshareddata/xcschemes/Hymns.xcscheme
```diff
@@ -78,6 +78,16 @@
...
       </BuildableProductRunnable>
+      <CommandLineArguments>
+         <CommandLineArgument
+            argument = "-FIRDebugEnabled"
+            isEnabled = "YES">
+         </CommandLineArgument>
+         <CommandLineArgument
+            argument = ""
+            isEnabled = "YES">
+         </CommandLineArgument>
```
skywalkerdude, line 89: Did you add an empty argument on accident?

Hymns.xcodeproj/project.pbxproj
```diff
@@ -816,7 +816,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "\"${PODS_ROOT}/Fabric/run\"\n";
+			shellScript = "\"${PODS_ROOT}/FirebaseCrashlytics/run\"\n";
```
skywalkerdude, line 819: I se "/path/to/pods/directory/FirebaseCrashlytics/upload-symbols" in https://firebase.google.com/docs/crashlytics/upgrade-sdk?platform=ios#the_run_and_upload-symbols_scripts_are_now_in_firebasecrashlytics. Do you know what that's for?

Hymns/AppDelegate.swift
```diff
@@ -1,15 +1,15 @@
...
+import FirebaseCrashlytics
 
 @UIApplicationMain
 class AppDelegate: UIResponder, UIApplicationDelegate {
 
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
         // Override point for customization after application launch.
         FirebaseApp.configure()
-        Fabric.sharedSDK().debug = true
+        Crashlytics.crashlytics()
```
skywalkerdude, line 12: Do we need this line? It looks like this just gets an instance of Crashlytics

finben:
@skywalkerdude For some reason I can't comment on your comments lol... Anyways
1. Yeah, added on accident lol just removed.
2. The symbols is basically to get the command line interface stuff. I didn't add it because I saw that you were not using it with Fabric. I only suggest we add it when we actually are planning to need or use any Firebase Crashlytics CLI tools. Otherwise no point in adding the upload symbols.
3. I think we need to initialize it for it to be active. However, I tried removing and then crashing the app and am waiting to see the results. If it works we can remove.

PR #87, opened 2020-04-26
Surround Display Hymn view with a pager that allows us to scroll between consecutive hymns

skywalkerdude:
Please try it out to see how well it works.

There are still 2 tasks that need to be done before this PR can be submitted.

1. There's a bug where the surrounding hymns will be added into the history store. For example, if you look up Hymn 45, hymns 44 and 46 will be added also, which shouldn't happen.
2. We need to write unit tests :)


finben:
Seems like you are having the same problem I was having but because of the extra binding you are getting wierder bugs. But basically what it looks like to me is happening is that before you even navigate from HomeViewModel -> PageView it already initializes PageView for the entire list of options. 

It's how navigationLinks work in SwiftUI and something I am trying to find out how to get around. Like lets say you have View A with a bunch of navigationLinks in a big list. Well it already initializes View B for each of the options before you even select which one you are going to. I was trying some stuff with initializing with onAppear() but it didn't really work out.

finben:
Do you still want to add this in? Is it worth adding even if it is a bit slow?

skywalkerdude:
probably not worth adding ;(

PR #95, opened 2020-04-28
expands the touchable space by moving to the HStack

finben:


Hymns/Common/SearchBar.swift
```diff
@@ -27,6 +24,11 @@ struct SearchBar: View {
                         self.searchText = ""
                     }
                 }, label: {Image(systemName: "xmark.circle.fill").opacity(self.searchText.isEmpty ? 0.0 : 1.0)})
+            }.onTapGesture {
+                if !self.searchActive {
+                    self.searchActive = true
```
skywalkerdude, line 29: What happens to the cancel button then if you do this?

Hymns/Common/SearchBar.swift
```diff
@@ -27,6 +24,11 @@ struct SearchBar: View {
...
+                }
+                print("Show details for Enterprise")
```
skywalkerdude, line 31: delete print statement?

Hymns/Common/SearchBar.swift
finben: It's fine. Their are two HStacks. Cancel button is part of the outter HStack which is not effected by this since this is on the inner stack

Hymns/Common/SearchBar.swift
finben: yeah lol..

finben:
So, I tried a bunch of stuff and nothing worked lol... I still recommend approving this PR since it at least gets us closer to what we want. It expands the selectable space all the way left including the magnifying class. And does at least bring the touchable space in closer on the right a little bit as well... I'm not sure what else to say then maybe the user if they click all the way on the right are just going to have to click again a little closer to the center lol...

PR #98, opened 2020-04-29
Work to get scrollview titles working as in spec

finben:
Biggest issue is truncation. There is simply no way to stop navigation bar titles from truncating. Drove me crazy today. I am still looking for options but if not we may have to do something more custom than using the navigation bar which would be unfortunate and probably much more difficult to build out.


skywalkerdude:
Isn't the nav title that we use already a custom component?

I agree that it seems infeasible to do with the stock NavigationTitle that SwiftUI provides. But I don't think it's that difficult to do if you listen to the drag gesture and shrink and position it based on the y-translation value

finben:
Closing branch to go with another approach

PR #106, opened 2020-05-01
Using geometry reader to work with scroll view title change

finben:
This is the third thing I am trying. I could not get drag gestures to work because the hymns are within a scrollview and the drag gestures don't work within scrollviews from what I could figure out. So this is an approach using geometry reader to measure when the Y axis has changed and display different titles based on that.


finben:
I have refactored this code to match our current code base much better and moved to branch scrollTitle. Closing this PR

PR #111, opened 2020-05-03
add and integrate grdb (sqlite database) wrapper

skywalkerdude:
Not done with unit tests yet, but will continue to write them after this commit


Hymns/Infrastructure/Data/HymnalApiService.swift
```diff
@@ -6,14 +6,11 @@ import Resolver
  * Service to contact the Hymnal API.
  */
 protocol HymnalApiService {
-    func getHymn(hymnType: HymnType, hymnNumber: String, queryParams: [String: String]?) -> AnyPublisher<Hymn, ErrorType>
+    func getHymn(_ hymnIdentifier: HymnIdentifier) -> AnyPublisher<Hymn, ErrorType>
```
finben, line 9: Is their an advantage to removing the label using _ ? Just wondering what made you decide to go this route, seems like it would decrease readability across the code base.

Hymns/Infrastructure/Data/HymnDataStore.swift
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Service to contact the Hymnal API.
+ */
+protocol HymnDataStore {
+    func saveHymn(_ entity: HymnEntity)
+    func getHymn(_ hymnIdentifier: HymnIdentifier) -> AnyPublisher<HymnEntity?, ErrorType>
+}
+
+/**
+ * Implementation of HymnalApiService that uses Alamofire.
```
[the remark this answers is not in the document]
skywalkerdude: sorry, leftover comment. fixed

Hymns/Common/SongResultView.swift
```diff
@@ -7,7 +7,7 @@ struct SongResultView: View {
     init(viewModel: SongResultViewModel) {
         self.viewModel = viewModel
     }
-
+// WTF COMMENT
```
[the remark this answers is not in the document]
skywalkerdude: haha oops. Removed! :P

Hymns/Infrastructure/Data/HymnalApiService.swift
skywalkerdude: From looking at how others do it, "_" is used to make things less verbose when it's clear what the argument is. So for example, for getHymn(...), it's clear from the method name that you are trying to get a hymn, so passing in the hymn identifier is intuitive without an explicit label? I dunno. What do you think, for this particular case?

skywalkerdude:
> Huge commit. So is the sheet music, guitar chords, etc now being retrieved and available to be implemented within our app? That is what it looks like to me.

You were always able to retrieve the sheet music and guitar chords and stuff. This just hooks up the local database so you can get those things without hitting the network

Hymns/Infrastructure/Data/HymnalApiService.swift
finben: Yeah, that makes sense. And I agree it is pretty intuitive here. I was just wondering.

PR #113, opened 2020-05-04
Large tab bar icons, weight fix, and padding fix

finben:
Addresses two issues from Emily on our trello board. She asked for larger tab bar icons, heavier weight on our titles, and padding to be the same for title as it is for search bar.


finben:
Yeah, it is like that for me as well. However, I just checked it on Johanna's iPhone and it was fine. I am wondering if it may be a simulator bug. Can you take a look on your actual device when you get a chance?

skywalkerdude:
It happens on my phone too
95907067_231696624790509_8840687830851125248_n

finben:
Looks like this bug was not introduced here. It is happening on our master branch :O

skywalkerdude:
yeah that's weird. Do you remember when you first noticed it? Maybe we can revert to older versions and see if it appears there to find out when it started appearing

skywalkerdude:
It could also be an xcode 11.4.1 bug. We can try installing the app from TestFlight and see if it appears there? If it doesn't appear there, then it's likely a non-issue

finben:
I am just going to get this merged in. It does not fix the issue however, it is also not the reason for the issue. We can address the bug in another PR. It is some sort of wierd bug that seems to have to do with the .hiddennavigationbar() in SwiftUI when used within a NavigationView(TabView)).

PR #115, opened 2020-05-04
Alter the cancel animation to address Emily's Trello Issue

finben:
Emily's Trello issue is the following:

 'Cancel' CTA '
Cancel' needs to already be at the right of the search bar when it comes in. Right now it's coming in from the right detached from the search bar.

### DIFF
--- a/Hymns/Common/CustomSearchTransition.swift
+++ b/Hymns/Common/CustomSearchTransition.swift
@@ -1,0 +1,9 @@
+import SwiftUI
+
+extension AnyTransition {
+    static var customSearchTransition: AnyTransition {
+        let insertion = AnyTransition.opacity
+        let removal = AnyTransition.move(edge: .trailing)
+        return .asymmetric(insertion: insertion, removal: removal)
+    }
+}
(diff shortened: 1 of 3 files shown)


Hymns/Common/CustomSearchTransition.swift
```diff
@@ -0,0 +1,9 @@
+import SwiftUI
+
+extension AnyTransition {
+    static var customSearchTransition: AnyTransition {
+        let insertion = AnyTransition.opacity
+        let removal = AnyTransition.move(edge: .trailing)
+        return .asymmetric(insertion: insertion, removal: removal)
+    }
```
skywalkerdude, line 8: is it necessary to have this, rather than just to do it inline? I don't foresee this specific animation having much of a reuse case.

Hymns/Common/SearchBar.swift
```diff
@@ -46,8 +46,8 @@ struct SearchBar: View {
                     }
                 }
                 .foregroundColor(Color(.systemBlue))
-                .transition(AnyTransition.move(edge: .trailing))
-                .animation(.easeInOut(duration: 0.2))
+                .transition(.customSearchTransition)
```
skywalkerdude, line 49: like can we just have this:

`.transition(.asymmetric(insertion: .opacity, removal: .move(edge: .trailing)))`

finben: Nice, yeah much better thanks!

PR #116, opened 2020-05-04
Changing text for dark mode colors on Figma

finben:
The only thing is to meet the Figma dark mode specs, I don't see a way to go around using custom asset catalog colors. There are certain things Emily is wanting that don't exactly match an Apple default color set that I can find by default. For instance to have Black text in normal mode but gray in dark mode.


finben:
Emily said that we can leave the placeholder text color the same and that it looks fine. I couldn't find an easy way to change it. This should get us looking much better in dark mode.

PR #118, opened 2020-05-05
A potential fix to the tab bar issue

finben:
I will admit this is a bit hacky and it does slow the launch a tad bit so may not be worth it. However, it does solve the problem lol.... Basically onAppear it selects the home tab and changes the state so that the bug gets resolved.


finben:
Actually, after running several time tests it does not slow down the launch at all. So it is indeed a viable solution now lol!

PR #120, opened 2020-05-06
Add chords, piano, and guitar tabs

finben:


Hymns/Display/DisplayHymnToolbarView.swift
```diff
@@ -0,0 +1,52 @@
...
+                }
+                Spacer()
+            }
+            Divider().edgesIgnoringSafeArea(.horizontal)
+            if self.toolbarTab == "lyrics" {
+                HymnLyricsView(viewModel: self.viewModel.hymnLyricsViewModel).padding(.horizontal)
+            } else if self.toolbarTab == "chords" {
+                // Text("\(self.viewModel.identifier.hymnNumber)")
+                // print("\(self.viewModel.identifier.hymnNumber)")
+                WebView(request: URLRequest(url: URL(string: "https://www.hymnal.net/en/hymn/\(self.viewModel.identifier.hymnType.abbreviatedValue)/\(self.viewModel.identifier.hymnNumber)/f=gtpdf")!))
```
skywalkerdude, line 38: So you have, in the hymnEntity object, the urls for each of the sheet music (both svg and pdf). I think it's probably better to use that than to hardcode the urls here

Hymns/Display/DisplayHymnToolbarView.swift
```diff
@@ -0,0 +1,52 @@
...
+            } else if self.toolbarTab == "guitar" {
+                WebView(request: URLRequest(url: URL(string: "https://www.hymnal.net/en/hymn/\(self.viewModel.identifier.hymnType.abbreviatedValue)/\(self.viewModel.identifier.hymnNumber)/f=pdf")!))
```
skywalkerdude, line 40: The loading is a little show here. We click it, and then it loads. Maybe this can be a follow-up, but how can we optimize this to maybe fetch the pdf/svg sheet music earlier or at the very least cache them after they are loaded so there won't be that long delay between when you click and when the images come up?

skywalkerdude:
I made a commit with something closer to what I had in mind. Let me know if that makes sense!

Hymns/Common/WebView.swift
```diff
@@ -6,13 +6,18 @@ import WebKit
...
+    let url: URL?
 
     func makeUIView(context: Context) -> WKWebView {
         return WKWebView()
     }
 
     func updateUIView(_ uiView: WKWebView, context: Context) {
-        uiView.load(request)
+        guard let url = url else {
+            // TODO show error state
```
skywalkerdude, line 17: Could you implement this error state?

Hymns/Display/DisplayHymnToolbarView.swift
```diff
@@ -0,0 +1,62 @@
+import SwiftUI
+
+struct DisplayHymnToolbarView: View {
+    @State var toolbarTab = "lyrics"
```
skywalkerdude, line 4: I think it might be good to make this an enum. That way in that little if/else clause, we can just use a switch statement and have compile-time assurance that it's exhaustive

Hymns/Display/DisplayHymnToolbarView.swift
```diff
@@ -0,0 +1,62 @@
...
+            } else if self.toolbarTab == "piano" && self.viewModel.pianoUrl != nil {
+                WebView(url: self.viewModel.pianoUrl!)
+            }
+        }
+    }
+}
+
+struct DisplayHymnToolbarView_Previews: PreviewProvider {
+    static var previews: some View {
+        DisplayHymnToolbarView(viewModel: DisplayHymnViewModel(hymnToDisplay: PreviewHymnIdentifiers.hymn40))
```
skywalkerdude, line 60: Could you make some previews here, so we can see what each state looks like without having to comment stuff out and run the code? That way we can catch visual bugs a little easier.

(i.e. only lyrics are present, lyrics + chords, lyrics + chords + guitar, lyrics + chords, + guitar + piano)

Hymns/Display/DisplayHymnViewModel.swift
```diff
@@ -59,7 +62,43 @@ class DisplayHymnViewModel: ObservableObject {
                     self.historyStore.storeRecentSong(hymnToStore: self.identifier, songTitle: title)
             }).store(in: &disposables)
     }
-
+    
+    func fetchHymnChords() {
```
skywalkerdude, line 66: could you unit test this method :D

Hymns/Display/DisplayHymnToolbarView.swift
```diff
@@ -0,0 +1,52 @@
...
+                    self.toolbarTab = "piano"
+                }
+                Spacer()
+            }
+            Divider().edgesIgnoringSafeArea(.horizontal)
+            if self.toolbarTab == "lyrics" {
+                HymnLyricsView(viewModel: self.viewModel.hymnLyricsViewModel).padding(.horizontal)
+            } else if self.toolbarTab == "chords" {
+                // Text("\(self.viewModel.identifier.hymnNumber)")
+                // print("\(self.viewModel.identifier.hymnNumber)")
```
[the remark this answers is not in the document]
finben: Resolved

Hymns/Display/DisplayHymnToolbarView.swift
finben: Outdated

Hymns/Display/DisplayHymnToolbarView.swift
finben: I updated to cache them at least they will only each load once now... Maybe on a follow up branch we can go further to fetch them earlier on.

Hymns/Common/WebView.swift
```diff
@@ -4,15 +4,36 @@ import WebKit
...
-    let request: URLRequest
+    static var cache = [URL: WKWebView]()
+
+    let url: URL?
 
     func makeUIView(context: Context) -> WKWebView {
-        return WKWebView()
+        guard let url = url else {
+            // TODO show error state
+             fatalError()
```
skywalkerdude, line 17: What does fatalError do? I don't think we want to crash the app if the url is empty... Maybe just showing an empty/error state is good enough. (maybe we can ask Emily to design one)

Hymns/Display/DisplayHymnToolbarView.swift
```diff
@@ -12,43 +12,43 @@ struct DisplayHymnToolbarView: View {
...
                 HymnLyricsView(viewModel: self.viewModel.hymnLyricsViewModel).padding(.horizontal)
-            } else if self.toolbarTab == "chords" && self.viewModel.chordsUrl != nil {
+            } else if self.toolbarTab == ToolbarTabs.chords && self.viewModel.chordsUrl != nil {
                 WebView(url: self.viewModel.chordsUrl!)
-            } else if self.toolbarTab == "guitar" && self.viewModel.guitarUrl != nil {
+            } else if self.toolbarTab == ToolbarTabs.guitar && self.viewModel.guitarUrl != nil {
                 WebView(url: self.viewModel.guitarUrl!)
-            } else if self.toolbarTab == "piano" && self.viewModel.pianoUrl != nil {
+            } else if self.toolbarTab == ToolbarTabs.piano && self.viewModel.pianoUrl != nil {
                 WebView(url: self.viewModel.pianoUrl!)
```
skywalkerdude: can you use a switch statement instead of if/else's?

Hymns/Display/DisplayHymnToolbarView.swift
finben: I couldn't figure this out :/ Whenever, I use the enum cases with the Webview it does not reload when toolbarTab refreshes. I tried some solutions from https://jasonzurita.com/swiftui-if-statement/ with the same results. Also, by using the enum we have to type erase the views since they are not all webviews (ake lyrics). By doing this I read we are also sacrificing some performance since SwiftUI does not know the view type upfront. Idk if this is also a consideration to not go the switch route. I did move the string to an enum though.

Hymns/Display/DisplayHymnViewModel.swift
finben: I wanted to check with you if we should try and combine this method into fetchHymn? What do you think would be better? If we leave it as is I will go forward with trying to write the unit tests?

Hymns/Display/DisplayHymnToolbarView.swift
finben: Doesn't work well in views :/ see my above comment on your enum comment. Won't refresh like if/else

Hymns/Common/WebView.swift
finben: Good point... Yeah that would be good.

Hymns/Display/DisplayHymnToolbarView.swift
finben: Added in recent commit

Hymns/Common/WebView.swift
finben: Not really sure what to do here... Doesn't it just return a blank now anyways?

Hymns/Display/DisplayHymnToolbarView.swift
```diff
@@ -15,28 +15,22 @@ struct DisplayHymnToolbarView: View {
...
                         Text("GUITAR").foregroundColor(self.toolbarTab == ToolbarTabs.guitar ? .accentColor : .primary).onTapGesture {
                             self.toolbarTab = ToolbarTabs.guitar
                         }
                     }
-                }
-                viewModel.chordsUrl.map { _ in
                     HStack {
                         Spacer()
                         Text("PIANO").foregroundColor(self.toolbarTab == ToolbarTabs.piano ? .accentColor : .primary).onTapGesture {
                             self.toolbarTab = ToolbarTabs.piano
```
skywalkerdude, line 33: What loading issue does this cause? They're necessary because you don't want to show a "guitar" option if there is no guitar music to show. I map over the optional so that if the optional is nil, then it won't show that item

Hymns/Display/DisplayHymnToolbarView.swift
skywalkerdude: Yeah, I think you are right. Let's keep it as if/else then!

Hymns/Common/WebView.swift
skywalkerdude: Yeah maybe just keep the TODO and we'll figure it out later

Hymns/Display/DisplayHymnToolbarView.swift
finben: Some of the HStack was coming in slow. So lyrics would immediately be there and then later the others would appear as they were mapped in. Do you do that on the Android app as well? Having the tabs just not show up if there is no music sheets?

skywalkerdude: yup, on Android I don't show the buttons if there is no music for that song. I wonder why they're coming slow. It didn't seem slow when I tried it

Hymns/Display/DisplayHymnViewModel.swift
skywalkerdude: Hmmm, good question. If I were doing it, I probably would have combined them. But now that I look at it, I think I rather like having two separate methods. It keeps things a lot cleaner. The only thing concern would be how it loads on the simulator/device.

Can you experiment a little bit? I tried turning the database off by changing this line to nil: (https://github.com/HymnalDreamTeam/Hymns-iOS/blob/master/Hymns/Infrastructure/Data/HymnDataStore.swift#L48) and looked at how it loaded.

It looks a little bit janky. The title/lyrics would load, then the lyrics bar would suddenly appear. If that's the case, it might be better to just combine them into one method

skywalkerdude: Just a tip for when you do this: feel free to get rid of the "map" step and do everything in the sink step. I don't think it makes sense to map the hymn to an individual component anymore, since we need multiple things from it.

Hymns/Display/DisplayHymnToolbarView.swift
finben: Okay, it is looking good now...

Hymns/Display/DisplayHymnViewModel.swift
finben: Okay, now I have added prefetching into the fetch hymn chords method. At this point, because it is prefetching and loading the webviews I think it makes more sense to have them as two methods. We definitely don't want the preloading of chords slowing down the other method?

skywalkerdude: I don't think prefetching the sheet music would slow down the hymn fetching at all. We aren't waiting on the prefetch to complete before displaying the lyrics. It should be something like this:

getHymn() -> receive hymn -> prefetch sheet music (asynchronously fire and forget that returns immediately) -> display lyrics

HymnsTests/Display/DisplayHymnViewModelSpec.swift
```diff
@@ -184,6 +184,60 @@ class DisplayHymnViewModelSpec: QuickSpec {
...
+                                it("Piano url should be /en/hymn/c/1151/f=ppdf") {
+                                    expect(target.chordsUrl).to(equal(URL(string: "/en/hymn/c/1151/f=ppdf")))
+                                }
+                                it("Guitar url should be /en/hymn/c/1151/f=pdf") {
+                                    expect(target.guitarUrl).to(equal(URL(string: "/en/hymn/c/1151/f=pdf")))
+                                }
+                                it("Text url should be /en/hymn/c/1151/f=gtpdf") {
+                                    expect(target.pianoUrl).to(equal(URL(string: "/en/hymn/c/1151/f=pdf")))
+                                }
+                            }
```
skywalkerdude, line 236: this looks right to me. What do you mean it's not doing anything?

Hymns/Display/DisplayHymnViewModel.swift
```diff
@@ -91,20 +91,29 @@ class DisplayHymnViewModel: ObservableObject {
                     self.chordsUrl = chordsPath.flatMap({ path -> URL? in
                         HymnalNet.url(path: path)
                     })
+                    if self.chordsUrl != nil {
+                        WebView.preload(url: self.chordsUrl!)
```
skywalkerdude, line 95: In your stack overflow link (https://stackoverflow.com/questions/48839180/preload-multiple-local-webviews), the WebViewPreloader there isn't static but is an object. If you follow that pattern, it'll be much easier to test/assert that you are preloading the correct thing here. Cuz currently, there's no way to assert this in your unit test

Hymns/Display/DisplayHymnViewModel.swift
```diff
@@ -91,20 +91,29 @@ class DisplayHymnViewModel: ObservableObject {
                     self.chordsUrl = chordsPath.flatMap({ path -> URL? in
                         HymnalNet.url(path: path)
                     })
+                    if self.chordsUrl != nil {
```
skywalkerdude, line 94: prefer optional unwrapping vs checking for nil: https://www.hackingwithswift.com/sixty/10/2/unwrapping-optionals

finben: Resolved in recent commit

Hymns/Display/DisplayHymnViewModel.swift
finben: Okay, I moved to a class. I tried to get away from using a static var but the problem is then the WebView representable struct complains about changing state in the view. So I have left it using a static var but made the functions associated with the class.

Hymns/Display/DisplayHymnViewModel.swift
finben: Check my recent commit... Hope I did this right. It seems to be working well for me..

Hymns/Display/DisplayHymnViewModel.swift
```diff
@@ -32,36 +40,64 @@ class DisplayHymnViewModel: ObservableObject {
...
+                guard let self = self else { return (nil, nil, nil) }
                 guard let hymn = hymn else {
-                    return nil
+                    return (nil, nil, nil)
                 }
-                return hymn.title.replacingOccurrences(of: "Hymn: ", with: "")
+                if self.identifier.hymnType == .classic {
+                    return ("Hymn \(self.identifier.hymnNumber)", hymn.lyrics, hymn.pdfSheet)
+                }
+                return (hymn.title.replacingOccurrences(of: "Hymn: ", with: ""), hymn.lyrics, hymn.pdfSheet)
```
skywalkerdude, line 59: I made one tiny tweak by removing this map, but other than that, it looks perfect. Great job!

HymnsTests/Display/DisplayHymnViewModelSpec.swift
```diff
@@ -135,6 +155,49 @@ class DisplayHymnViewModelSpec: QuickSpec {
...
                                     }
+                                    it("Piano url should be http://www.hymnal.net/en/hymn/c/145/f=ppdf") {
+                                        expect(target.chordsUrl).to(equal(URL(string: "http://www.hymnal.net/en/hymn/c/1151/f=gtpdf")))
+                                    }
+                                    it("Guitar url should be http://www.hymnal.net/en/hymn/c/145/f=pdf") {
+                                        expect(target.guitarUrl).to(equal(URL(string: "http://www.hymnal.net/en/hymn/c/1151/f=pdf")))
+                                    }
+                                    it("Text url should be http://www.hymnal.net/en/hymn/c/145/f=gtpdf") {
+                                        expect(target.pianoUrl).to(equal(URL(string: "http://www.hymnal.net/en/hymn/c/1151/f=ppdf")))
+                                    }
```
skywalkerdude, line 166: FYI, I generally like making the expected url a constant and then just pass it in to the assertions. That way, you can avoid errors like this, where you say "**text** url should be ...c/**145**/..." but you're asserting on **pianoUrl** being "...c/**1151**/..."

PR #122, opened 2020-05-07
Fix firebase bundle

finben:
I did not really fully setup the debug because I was not sure if the current DEBUG file is different. But this is what it looks like with the Beta setup. Look under the build phase google plist setup script.

Also, if you look under our firebase directory you will notice that we basically need to add a new plist associated with different firebase projects for each app product.


Hymns/Firebase/Dev/GoogleService-Info.plist
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
skywalkerdude, line 1: This seems to be the same thing as Debug? I think we can remove this

Hymns.xcodeproj/project.pbxproj
```diff
@@ -901,7 +901,7 @@
...
...vice-Info.plist found. Please ensure it's in the proper directory.\"\n    exit 1\nfi\n\n# Make sure the beta version of GoogleService-Info.plist exists\necho \"Looking for ${GOOGLESERVICE_INFO_PLIST} in ${GOOGLESERVICE_INFO_BETA}\"\nif [ ! -f $GOOGLESERVICE_INFO_BETA ]\nthen\n    echo \"No Beta GoogleService-Info.plist found. Please ensure it's in the proper directory.\"\n    exit 1\nfi\n\n# Get a reference to the destination location for the GoogleService-Info.plist\nPLIST_DESTINATION=${BUILT_PRODUCTS_DIR}/${PRODUCT_NAME}.app\necho \"Will copy ${GOOGLESERVICE_INFO_PLIST} to final destination: ${PLIST_DESTINATION}\"\n\n# Copy over the prod GoogleService-Info.plist for Release builds\nif [ \"${CONFIGURATION}\" == \"Release\" ]\nthen\n    echo \"Using ${GOOGLESERVICE_INFO_PROD}\"\n    cp \"${GOOGLESERVICE_INFO_PROD}\" \"${PLIST_DESTINATION}\"\n\n# Copy over the beta GoogleService-Info.plist for Beta builds\nelif [ \"${CONFIGURATION}\" == \"Beta\" ]\nthen\n    echo \"Using ${GOOGLESERVICE_INFO_BETA}\"\n    cp \"${GOOGLESERVICE_INFO_BETA}\" \"${PLIST_DESTINATION}\"\nelse\n    echo \"Using ${GOOGLESERVICE_INFO_DEV}\"\n    cp \"${GOOGLESERVICE_INFO_DEV}\" \"${PLIST_DESTINATION}\"\nfi\n";
```
skywalkerdude, line 904: What did you change here?

I'm thinking for the time being, we don't need a Firebase Config for the Beta version. I don't think we'll be launching Beta to anybody, since the app store requires the release build and doesn't take a Beta distribution. (The Beta distribution channel is for the enterprise level, apparently, which we aren't pay for.

I'm thinking about actually removing the Beta build actually and just having a release and debug.

I think if you revert this, and remove Hymns/Firebase/Dev/GoogleService-Info.plist, we should be good to go, since I already replaced the /Debug/Google-Service-Info.plist

skywalkerdude:
Also a bit concerned that swiftlint isn't running. Can you add a dummy comment somewhere in the code to make sure swift lint runs?

Hymns.xcodeproj/project.pbxproj
finben: Okay, I am just going to close down this branch.

PR #131, opened 2020-05-11
adding /path/to/pods/directory/FirebaseCrashlytics/upload-symbols to …

finben:
…end of build phase

per https://firebase.google.com/docs/crashlytics/upgrade-sdk

### DIFF
--- a/Hymns.xcworkspace/xcshareddata/WorkspaceSettings.xcsettings
+++ b/Hymns.xcworkspace/xcshareddata/WorkspaceSettings.xcsettings
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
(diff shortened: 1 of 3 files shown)


finben:
This doesn't work but it is how they said to do it in the documentation. I tried some other things as well but I am wondering if Crashlytics initialization is having trouble finding our PList file since it is not in the project directory anymore. 

See below link some solutions by agiguere  where they are using if elif to then go get the googleplist file in its location based on build.
https://github.com/firebase/firebase-ios-sdk/issues/3299

Hymns.xcworkspace/xcshareddata/WorkspaceSettings.xcsettings
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>PreviewsEnabled</key>
+	<false/>
+</dict>
+</plist>
```
skywalkerdude, line 8: What is this?

PR #138, opened 2020-05-11
Blue line indicator under hymn lyrics bar

finben:
My middle of the night work. I just don't know how to pass variables into the enum content to make it dynamic? How can I get something like a viewModel or the URL's to be usable for the enum content. 

Once this is fixed though, I think it looks pretty good repurposing your old tab bar work for this usage. I guess your custom tab bar work won't be in vain after all ;)


skywalkerdude:
I don't think it's a bug with SwiftUI. It has to do with this line in WebView:

```        if uiView.url == nil {
            uiView.load(URLRequest(url: url))
        }```

If you just call uiView.load(...) every time, without the if check, it works like a charm.

**Comment by skywalkerdude** [comment]
Also, could you fix the code to not essentially copy and paste the entire tab view indicator class? You can make HymnLyricsTab implement TabItem and essentially use the indicator class out of the box

**Comment by finben** [comment]
I went ahead and merged your branch. I still think we should just go with the if statements though... If you checkout commit 713d480 you will see how much smoother the tabbing is when it doesn't have to update uiview but can just smoothly switch between. Unfortunately, I haven't found another way to do this besides the if statements. It doesn't work with enum switch case. Idk, we can look into the html caching but from what I have read people complain that even calling uiView.load(html) still takes a long time since it is making the webview and that takes the longest time.

File: Hymns/Display/DisplayHymnView.swift [hunk 1]
```diff
@@ -38,73 +27,51 @@ struct DisplayHymnView: View {
...
-        classic40ViewModel.lyrics = [Verse]()
-        classic40ViewModel.chordsUrl = URL(string: "http://www.hymnal.net/en/hymn/h/40/f=gtpdf")
-        classic40ViewModel.guitarUrl = URL(string: "http://www.hymnal.net/en/hymn/h/40/f=pdf")
-        classic40ViewModel.pianoUrl = URL(string: "http://www.hymnal.net/en/hymn/h/40/f=ppdf")
         let classic40Lyrics = HymnLyricsViewModel(hymnToDisplay: PreviewHymnIdentifiers.hymn40)
-        classic40Lyrics.lyrics = [VerseViewModel(verseNumber: "1", verseLines: classic40_preview.lyrics[0].verseContent),
-                                  VerseViewModel(verseLines: classic40_preview.lyrics[1].verseContent),
-                                  VerseViewModel(verseNumber: "2", verseLines: classic40_preview.lyrics[2].verseContent),
-                                  VerseViewModel(verseNumber: "3", verseLines: classic40_preview.lyrics[3].verseContent),
-                                  VerseViewModel(verseNumber: "4", verseLines: classic40_preview.lyrics[4].verseContent)]
```
skywalkerdude, line 84: I think we need these? Otherwise all the previews will have just the empty loading screen

Hymns/Display/DisplayHymnView.swift
```diff
@@ -38,73 +27,51 @@ struct DisplayHymnView: View {
...
         let classic1151ViewModel = DisplayHymnViewModel(hymnToDisplay: PreviewHymnIdentifiers.hymn1151)
         classic1151ViewModel.title = "Hymn 1151"
         classic1151ViewModel.isFavorited = false
         let classic1151Lyrics = HymnLyricsViewModel(hymnToDisplay: PreviewHymnIdentifiers.hymn1151)
-        classic1151Lyrics.lyrics = [VerseViewModel(verseNumber: "1", verseLines: classic1151_preview.lyrics[0].verseContent),
-                                    VerseViewModel(verseLines: classic1151_preview.lyrics[1].verseContent),
-                                    VerseViewModel(verseNumber: "2", verseLines: classic1151_preview.lyrics[2].verseContent),
-                                    VerseViewModel(verseNumber: "3", verseLines: classic1151_preview.lyrics[3].verseContent),
-                                    VerseViewModel(verseNumber: "4", verseLines: classic1151_preview.lyrics[4].verseContent)]
-        classic1151ViewModel.hymnLyricsViewModel = classic1151Lyrics
```
skywalkerdude, line 97: ditto

Hymns/Display/DisplayHymnView.swift
```diff
@@ -38,73 +27,51 @@ struct DisplayHymnView: View {
...
+            .piano(WebView(url: URL(string: "http://www.hymnal.net/en/hymn/h/1151/f=ppdf")).eraseToAnyView())]
+
+        let classic1151 = DisplayHymnView(viewModel: classic1151ViewModel)
         let classic1334ViewModel = DisplayHymnViewModel(hymnToDisplay: PreviewHymnIdentifiers.hymn1334)
         classic1334ViewModel.title = "Hymn 1334"
         classic1334ViewModel.isFavorited = true
         let classic1334Lyrics = HymnLyricsViewModel(hymnToDisplay: PreviewHymnIdentifiers.hymn1334)
-        classic1334Lyrics.lyrics
-            = [VerseViewModel(verseNumber: "1", verseLines: classic1334_preview.lyrics[0].verseContent)
-        ]
```
skywalkerdude, line 106: ditto

skywalkerdude:
> I went ahead and merged your branch. I still think we should just go with the if statements though... If you checkout commit [713d480](https://github.com/HymnalDreamTeam/Hymns-iOS/commit/713d480da39f278475daaf31d7a138fa3483c7eb) you will see how much smoother the tabbing is when it doesn't have to update uiview but can just smoothly switch between. Unfortunately, I haven't found another way to do this besides the if statements. It doesn't work with enum switch case. Idk, we can look into the html caching but from what I have read people complain that even calling uiView.load(html) still takes a long time since it is making the webview and that takes the longest time.

I'm not sure what you're talking about. The commit you linked me only removes a Spacer().

Could you put in the if statement as another PR that's based off of this one (kind of like what I did with toolbar2) and we can take a look

skywalkerdude:
I think you also need to fix the unit tests because DisplayHymnViewModel changed

finben:
The article you sent me was excellent. It really helped to understand what was actually going on here. So the reason my toolbar fast branch was tabbing smoothly was essentially because by only using a bunch of if statements makeUIView was just always getting called  again and again instead of it being in the updateUIView phase. For more information on the lifecycle phases there is https://sarunw.com/posts/uikit-in-swiftui/ which is very helpful. 

However, when we use the enum to return WebViews instead of makeUIView each time we tab it is only called the first time to initialize the webview and everytime we are tabbing we are in the updateUIView phase and only updateUIView is being called. Which usually and theoretically this makes sense and is intended functionality of how SwiftUI integrates UIKit. 

However, we are somewhat trapped by this in using our cached webviews. Because in updateUIView we are not able to return anything except that original webview which can not be mutated or substituted. We can only change it's state. But in this particular situation what we really want to do is just return another webview entirely when we tab over that would come from our cache. 

So, I have a dumb hack that circumvents us from ever calling updateUIView and instead each time we tab makeUIView will be called. Lol! 

PS. I did look into using a Coordinator but the problem I was finding is that essentially they are used to have changes in SwiftUI state or binding variables etc. retrigger updateUIView. But that still seems to leave us stuck with using updateUIView which cannot return our cached webviews. And instead would need to load them each time.

we switched tabs which immediately could return a cached Wkwebview. s

The

skywalkerdude:
Thanks for the detailed explanation. I did some research and it looks like there's this thing called PDFView, which works much better. This seems like a more viable solution ;)

Hymns/Infrastructure/Data/PdfPreloader.swift
```diff
@@ -0,0 +1,38 @@
...
+import Foundation
+import PDFKit
+import Resolver
+
+/**
+ * Asynchronously loads and caches PDF documents loaded from web URLs
+ */
+class PDFLoader {
+
+    private let analytics: AnalyticsLogger
```
skywalkerdude, line 11: oops, I left the analytics object in. You can remove that or add some analytics if you want to learn how to do that. Just followed the pattern in AnalyticsLogger

Hymns/Display/DisplayHymnView.swift
skywalkerdude: I fixed this for you, but can you fix the unit tests?

Hymns/Infrastructure/Analytics/AnalyticsLogger.swift
```diff
@@ -24,6 +24,10 @@ class AnalyticsLogger {
         ])
     }
 
+    func logDisplayMusicPDF(url: URL) {
+        Analytics.logEvent(DisplayMusicPDF.name, parameters: [ DisplayMusicPDF.Params.display_musicPDF.rawValue: url])
```
skywalkerdude, line 28: Did you check to see how this appears on Firebase's DebugView? I'm not sure it knows how to log a URL object

Hymns/Infrastructure/Analytics/EventType.swift
```diff
@@ -42,6 +42,18 @@ struct DisplaySong: AnalyticsEvent {
     // swiftlint:enable identifier_name
 }
 
+struct DisplayMusicPDF: AnalyticsEvent {
+
+    static let name = "display_hymnMusic_pdfs"
```
skywalkerdude, line 47: how come everything is snack case but hymnMusic is suddenly camelcase?

Hymns/Infrastructure/Data/PdfPreloader.swift
```diff
@@ -33,6 +33,7 @@ class PDFLoader {
     }
 
     func get(url: URL) -> PDFDocument? {
+        analytics.logDisplayMusicPDF(url: url)
```
skywalkerdude, line 36: If you are logging displaying the pdf, I think it makes more sense for it to be in the PDF Viewer itself, rather than in the cache. In the future, getting something from the cache,  doesn't necessarily mean you're going to immediately be displaying the pdf

skywalkerdude:
could you fix up the unit tests and we can get this in?

finben:
@skywalkerdude Want to help me debug why DisplayHymnViewModelSpec is not working? I'm kind of stuck :/ 

One error is that realm is being called from the wrong thread
The other is that in the nil repository results test getHymn is not being called. " Error: Got 0 invocations of 'getHymn("

Hymns/Infrastructure/Data/PDFPreloader.swift
```diff
@@ -33,3 +39,11 @@ class PDFLoader {
         return cache[url]
     }
 }
+
+extension Resolver {
+    public static func registerPDFLoader() {
+        register(PDFLoader.self) {
+            return PDFLoaderImpl() as PDFLoader
```
skywalkerdude, line 46: you don't really need the return statement if it's just one-liine

Hymns/Infrastructure/Dependency Injection/AppDelegate+Injection.swift
```diff
@@ -23,7 +23,8 @@ extension Resolver: ResolverRegistering {
         register {URLSession.shared}.scope(application)
         register {AnalyticsLogger()}
         register {SystemUtil()}
-        register {PDFLoader()}
+        register {PDFLoaderImpl()}
```
skywalkerdude, line 26: You don't need this if you have registerPDFLoader()

HymnsTests/Display/DisplayHymnViewModelSpec.swift
```diff
@@ -5,26 +5,27 @@ import Quick
 @testable import Hymns
 
 class DisplayHymnViewModelSpec: QuickSpec {
-
+
```
skywalkerdude, line 8: pls fix

HymnsTests/Display/DisplayHymnViewModelSpec.swift
```diff
@@ -5,26 +5,27 @@ import Quick
...
-                webViewPreloader = mock(WebViewPreloader.self)
+                analytics = mock(AnalyticsLogger.self)
+                pdfLoader = mock(PDFLoader.self)
             }
             describe("fetching hymn") {
                 context("with nil repository result") {
                     beforeEach {
-                        target = DisplayHymnViewModel(backgroundQueue: testQueue, hymnToDisplay: classic1151, hymnsRepository: hymnsRepository,
-                                                      favoritesStore: favoritesStore, historyStore: historyStore)
+                        target = DisplayHymnViewModel(analytics: analytics, backgroundQueue: testQueue, favoritesStore: favoritesStore, hymnToDisplay: classic1151, historyStore: historyStore, pdfPreloader: pdfLoader)
```
skywalkerdude, line 28: Two things here:

1) You don't really need to mock the analytics logger because you never make any assertions on it, nor do you mock any of its methods. If you want to include it, then you should put assertions on it.
2) Noticed that there used to be a hymnsRepository: hymnsRepository in the constructor that you removed. That's the source of a lot of your test failures, since you are no longer using the mocked one and you're using the real one.

A big giveaway was this test failing:
```                       
it("should call hymnsRepository.getHymn") {
  verify(hymnsRepository.getHymn(classic1151)).wasCalled(exactly(1))
}
```

Because in fetchHymn, one of the first things we do is repository.getHymn:
```   
 func fetchHymn() {
        analytics.logDisplaySong(hymnIdentifier: identifier)
        repository
            .getHymn(identifier)
  ...
  }
```
So if `hymnsRepository.getHymn` is not being called, that should be a red flag that you're using the right repository.

HymnsTests/Display/DisplayHymnViewModelSpec.swift
skywalkerdude, line 8: pls fix

HymnsTests/Display/DisplayHymnViewModelSpec.swift
```diff
@@ -100,38 +89,25 @@ class DisplayHymnViewModelSpec: QuickSpec {
...
                                 }
                             }
                         }
                     }
                     context("for new song 145") {
                         beforeEach {
-                            target = DisplayHymnViewModel(backgroundQueue: testQueue, hymnToDisplay: newSong145, hymnsRepository: hymnsRepository,
-                                                          mainQueue: testQueue, favoritesStore: favoritesStore, historyStore: historyStore,
-                                                          webviewCache: webViewPreloader)
+                            target = DisplayHymnViewModel(analytics: analytics, backgroundQueue: testQueue, favoritesStore: favoritesStore, hymnToDisplay: newSong145, hymnsRepository: hymnsRepository,
```
skywalkerdude, line 109: Here you're missing the HistoryStore.

The raywenderlich article you sent me shows how to run these tests one at a time. https://www.raywenderlich.com/135-behavior-driven-testing-tutorial-for-ios-with-quick-nimble#toc-anchor-007
You can usually use that to isolate which tests are throwing the errors so you can figure out which tests to focus on.

HymnsTests/Display/DisplayHymnViewModelSpec.swift
```diff
@@ -36,20 +37,8 @@ class DisplayHymnViewModelSpec: QuickSpec {
...
-                        }
-                        it("chords should be nil") {
-                            expect(target.chordsUrl).to(beNil())
-                        }
-                        it("guitar url should be nil") {
-                            expect(target.guitarUrl).to(beNil())
-                        }
-                        it("piano url should be nil") {
-                            expect(target.pianoUrl).to(beNil())
-                        }
```
skywalkerdude, line 50: I don't know if you're just not done with the tests yet, but after deleting these cases, it'd probably be better to add new cases testing the new behavior.

Hymns/Infrastructure/Dependency Injection/AppDelegate+Injection.swift
```diff
@@ -20,14 +20,16 @@ extension Resolver: ResolverRegistering {
             encoder.nonConformingFloatEncodingStrategy = .throw
             return encoder
         })
+        register {URLSession.shared}.scope(application)
```
skywalkerdude, line 23: don't think you need this?

Hymns/Infrastructure/Data/PDFPreloader.swift
finben: Thanks!

HymnsTests/Display/DisplayHymnViewModelSpec.swift
finben: Nice! Thank you!

HymnsTests/Display/DisplayHymnViewModelSpec.swift
finben: I guess I deleted History store and the HymnsRepository without realizing it. Thanks...

HymnsTests/Display/DisplayHymnViewModelSpec.swift
finben: Added a new test verifying the number of tabs. Is there any other functionality or things you think we should test?

Hymns/Infrastructure/Dependency Injection/AppDelegate+Injection.swift
finben: Addressed in recent commit

HymnsTests/Display/DisplayHymnViewModelSpec.swift
skywalkerdude: I don't think we made that big of a change to the view model, so there's not that much to change.

But I do think there's some value in testing the ordering of the tabs, since that affects the order the tabs appear on the screen. Could you add those to the "should have x tabs" assertions too?

You can use the id to perform the assertion. So like 

```
expect(target.tabItems[0].id).to(equal("Lyrics"))
expect(target.tabItems[1].id).to(equal("Chrods"))
```
etc

Other than that, everything else looks good!

finben:
Thanks Luke, I think we are good lets get it merged to master?

PR #145, opened 2020-05-13
Audio player

finben:
Bottom sheet resource is https://swiftwithmajid.com/2019/12/11/building-bottom-sheet-in-swiftui/

Audio player main resource is https://github.com/ChrisMash/AVPlayer-SwiftUI

For some reason the combine stuff isn't working with our URLS that aren't straight up mp3 urls for example http://www.hymnal.net/en/hymn/h/894/f=mp3 that is coming from musicJson. However, the combine works when the url is a direct mp3 url such as https://www.hymnal.net/Hymns/NewSongs/mp3/ns0767.mp3

Need to figure that out more. Just committing what I have for now.


finben:
Update: I have been unable to get duration or any other meta data from the song when retrieving from the url saved in musicJSON. How did you go about this on the droid app?

skywalkerdude:
what do you mean duration? You can hit the urls directly to see what data you have inside the metadata: http://hymnalnetapi.herokuapp.com/v2/hymn/h/1151. Some of the other metadata keys are also in the MetaDataName enum: https://github.com/HymnalDreamTeam/Hymns-iOS/blob/master/Hymns/Infrastructure/Data/Models/MetaDatum.swift#L11

finben:
None of that really works. Both the metadata hitting the url directly as well as the MetaDataName enum do not contain what I am looking for which is song duration. Our MetaDataName mainly just contains two other urls one for medi and the other for downloading a mid file.

skywalkerdude:
yeah, i don't think song duration is really something we have access to at the moment. Why do you need it?

finben:
In my audio viewer it has a slider that goes along as the song plays. But it needs the length of the song to measure off of. So currently the music slider is always at the final position when playing a song

PR #151, opened 2020-05-18
Implement local search

skywalkerdude:


Hymns/Infrastructure/Analytics/AnalyticsLogger.swift
```diff
@@ -1,33 +1,48 @@
...
 /**
  * Wrapper for `FirebaseAnalytics` to help keep track of what we are logging and analyzing.
  */
 class AnalyticsLogger {
 
+    private let backgroundThread: DispatchQueue
+
+    init(backgroundThread: DispatchQueue = Resolver.resolve(name: "background")) {
+        self.backgroundThread = backgroundThread
+    }
```
finben, line 14: Why do you use custom dispatchQueues for background instead of the DispatchQueue.global? I am guessing it is better for dependency injection?

HymnsTests/Repository/HymnsRepositoryImpl_dbInitialized_dbMissTests.swift
```diff
@@ -3,7 +3,6 @@ import Mockingbird
 import XCTest
 @testable import Hymns
 
-// TODO delete later
```
finben, line 6: Should we delete now?

Hymns/Infrastructure/Analytics/AnalyticsLogger.swift
skywalkerdude: I don't use a custom dispatch queue. If you look at https://github.com/HymnalDreamTeam/Hymns-iOS/blob/master/Hymns/Infrastructure/Dependency%20Injection/AppDelegate%2BInjection.swift#L10, I'm merely injecting the DispatchQueue.global into here, so it is using the global dispatch queue

skywalkerdude:
> Big PR. Hard for me to thoroughly ensure quality... lol

yeah... I tried to split up the NetworkBoundPublisher stuff into its own PR, but apart from that, it just touches a lot of pieces

Issue #152, opened 2020-05-19
Fix .gitignore based on ikesyo's comment in https://github.com/HymnalDreamTeam/Hymns-iOS/pull/40#issuecomment-628811839_


skywalkerdude:
The reason for the problem is https://github.com/Quick/Nimble/tree/8.x-branch/Carthage/Checkouts/CwlPreconditionTesting/Dependencies is ignored due to https://github.com/HymnalDreamTeam/Hymns-iOS/blob/f1363ad71ee2a9ce47dcbe34d3b372762fb6dbef/.gitignore#L69-L70. I recommend you to change the entry to `/Dependencies/`.

_Originally posted by @ikesyo in https://github.com/HymnalDreamTeam/Hymns-iOS/pull/40#issuecomment-628811839_

ikesyo:
#66 should already resolved the issue.

PR #66, opened 2020-04-20
Remove pods

finben:

PR #153, opened 2020-05-19
Fix .gitignore based on ikesyo's comment

skywalkerdude:
Hopefully this makes the build faster. Closes https://github.com/HymnalDreamTeam/Hymns-iOS/issues/152

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,8 +67,8 @@ Pods/*
 Carthage/Build/
 
 # Accio dependency management
-# Dependencies/
-# .accio/
+Dependencies/
+.accio/
 
 # fastlane
 #
--- a/.gitignore
+++ b/.gitignore
@@ -67,8 +67,8 @@ Pods/*
 Carthage/Build/
 
 # Accio dependency management
-Dependencies/
-.accio/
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -67,8 +67,8 @@ Pods/*
 Carthage/Build/
 
 # Accio dependency management
-Dependencies/
-.accio/
+# Dependencies/
```
ikesyo, line 70: Looks like now that you ignore entire `Pods` dir (since #66) so this is meaningless.

skywalkerdude: @ikesyo, I ignore the entire Pods directory but I also do a `pod install` during the build step which will pull in all the pods. I can remove this and see if a simple pod install will work, but my impression was that only doing pod install will cause the build issue (https://github.com/Quick/Nimble/issues/490). I can revert this change and see if it still the case

skywalkerdude: ah, looks like it passes without the pod deintegrate. great!

ikesyo: > it passes without the pod deintegrate

Yes, that is the poiint.

PR #160, opened 2020-05-20
Actually use our preloader get function. Big speed boost switching tabs

finben:


Podfile.lock
```diff
@@ -84,14 +84,14 @@ PODS:
...
-  - Realm (4.4.1):
-    - Realm/Headers (= 4.4.1)
-  - Realm/Headers (4.4.1)
-  - RealmSwift (4.4.1):
-    - Realm (= 4.4.1)
+  - Realm (5.0.0):
+    - Realm/Headers (= 5.0.0)
+  - Realm/Headers (5.0.0)
+  - RealmSwift (5.0.0):
+    - Realm (= 5.0.0)
```
skywalkerdude, line 94: Is there a reason you had to run a pod update? I'm not against it, just curious why you did it.

Hymns/Common/PDFViewer.swift
```diff
@@ -3,7 +3,7 @@ import SwiftUI
 import PDFKit
 
 struct PDFViewer: UIViewRepresentable {
-
+    private let preloader: PDFLoader = Resolver.resolve()
```
skywalkerdude, line 6: lol, didn't realize we weren't using it

Podfile.lock
finben: I kept getting that realm error when I was running the project. This fixed it on my local by upgrading our realm version.

skywalkerdude: Could you be more specific on what "that realm error" is?

PR #162, opened 2020-05-21
Dynamically changing scroll title according to Emily's spec for lyrics

finben:


finben:
Sort of stuck at this point... I have no idea why this does not work? The published var is being updated from the scrollview geometry reader and the display hymn view and toolbar view both see that but they aren't updating their views. I'm not sure maybe, this is a case where I need to be using combine. 

It works fine when I use an environment object for the first hymn you do it on but then as soon as you try another hymn it crashes. So I think it will have to be done without using a singleton in the environment.

PR #165, opened 2020-05-21
Add beginnings of the display hymn bottom bar

skywalkerdude:
Simulator Screen Shot - iPhone 8 - 2020-05-21 at 01 01 26


skywalkerdude:
Yeah that's a good point. The Android one also can have up to 7 buttons, just most hymns only have 6 based on the data there is.

Could you bring this up to Emily?

Issue #163, opened 2020-05-21
Create empty state for favorites view according to figma


skywalkerdude:
<img width="549" alt="Screen Shot 2020-05-20 at 9 07 57 PM" src="https://user-images.githubusercontent.com/1427524/82522345-fcec0e80-9add-11ea-818e-3e5dd824a0cd.png">

Issue #164, opened 2020-05-21
Create empty state for home screen according to figma


skywalkerdude:
<img width="524" alt="Screen Shot 2020-05-20 at 9 11 31 PM" src="https://user-images.githubusercontent.com/1427524/82522498-7e43a100-9ade-11ea-964b-f5c94338d2f4.png">

PR #171, opened 2020-05-21
Adding cool images for empty views based on figma

finben:
Since I'm struggling with all my other branches.... I did something easy. lol


skywalkerdude:
Closes https://github.com/HymnalDreamTeam/Hymns-iOS/issues/163 and https://github.com/HymnalDreamTeam/Hymns-iOS/issues/164

Hymns/Assets.xcassets/favorites illustration.imageset/Contents.json
```diff
@@ -0,0 +1,21 @@
+{
+  "images" : [
+    {
+      "filename" : "favorites illustration.png",
```
skywalkerdude, line 4: Can you rename to Favorites Empty State or Empty Favorites Illustration or something that indicates it's the empty state rather than just "illustration"

Hymns/Assets.xcassets/search illustration.imageset/Contents.json
```diff
@@ -0,0 +1,21 @@
+{
+  "images" : [
+    {
+      "filename" : "search illustration.png",
```
skywalkerdude, line 4: ditto

Hymns/Favorites/FavoritesView.swift
```diff
@@ -14,7 +14,10 @@ struct FavoritesView: View {
         VStack {
             CustomTitle(title: "Favorites")
             if self.viewModel.favorites.isEmpty {
-                Text("Tap the ♡ on any hymn to add as a favorite").maxSize()
+                ZStack {
+                    Image("favorites illustration").offset(y: -70)
+                    Text("Tap the ♡ on any hymn to add as a favorite").maxSize().offset(y: 70)
```
skywalkerdude, line 19: I think the figma has a different string right

skywalkerdude, line 19: How come you didn't just use a VStack?

Hymns/Home/HomeView.swift
```diff
@@ -32,13 +32,23 @@ struct HomeView: View {
...
+                if viewModel.songResults.isEmpty {
+                    Spacer()
+                    HStack {
+                        Spacer()
+                        Image("search illustration")
+                        Spacer()
                     }
-                }.resignKeyboardOnDragGesture()
+                    Spacer()
+                } else {
```
skywalkerdude, line 43: I think there are probably better ways to center an image other than using Spacers and HStacks

Hymns/Home/HomeView.swift
```diff
@@ -32,13 +32,23 @@ struct HomeView: View {
             } else if viewModel.state == .empty {
                 Text("Did not find any songs matching:\n\"\(viewModel.searchParameter)\".\nPlease try a different request").padding().multilineTextAlignment(.center).maxSize(alignment: .center)
             } else {
-                List(viewModel.songResults) { songResult in
-                    NavigationLink(destination: songResult.destinationView) {
-                        SongResultView(viewModel: songResult)
-                    }.onAppear {
-                        self.viewModel.loadMore(at: songResult)
+                if viewModel.songResults.isEmpty {
```
skywalkerdude, line 35: Can we also hide the "Recent songs" label if it's empty?

finben: Addressed in recent commit

Hymns/Home/HomeView.swift
finben: So the reason for this is because the alignment left we have on the parent view keeps overriding this views alignment. I tried two other options of putting the Image alignment to center but it still ends up aligned left. Also, I tried .position(x: UIScreen.main.bounds.width/2) but that was also never in the right position. The one other thing we could do is geometry reader but that is probably going to be uglier than the HStack with spacers... Lmk what you think?

Hymns/Favorites/FavoritesView.swift
finben: VStack was tough to get aligned in the right position. The Image is always way to high and text way to low. I could force them into only taking up so much space but thought that might not be too good for various devices. This ZStack route gave me what I want on various device sizes every time even when testing on iPad so it seems like as good of a way as any to me.

Hymns/Assets.xcassets/favorites illustration.imageset/Contents.json
finben: Resolved in recent commit

Hymns/Assets.xcassets/search illustration.imageset/Contents.json
finben: Resolved in recent commit

Hymns/Favorites/FavoritesView.swift
finben: Resolved in recent commit

Hymns/Assets.xcassets/favorites illustration.imageset/Contents.json
skywalkerdude: You didn't actually commit anything

Hymns/Home/HomeView.swift
skywalkerdude: What if you just did Image(...).maxSize()?

Hymns/Favorites/FavoritesView.swift
skywalkerdude: what about

```                VStack {
                    Image("favorites illustration")
                    Text("Tap the ♡ on any hymn to add as a favorite")
                }.maxSize()```

File: Hymns/Home/HomeView.swift (hunk 4 above)
Comment (finben) [reply Hymns/Home/HomeView.swift]: Lol! Nice... I didn't think about that. That works pretty well here but the only problem is when it is in landscape mode. It runs up into the search bar to reach it's max size. In regular orientation this would have worked well though.

File: Hymns/Favorites/FavoritesView.swift (hunk 3 above)
Comment (finben) [reply Hymns/Favorites/FavoritesView.swift]: That works pretty well actually. I went ahead and committed it. At first it was too low for the spec but once an offset is added it looks good.

File: Hymns/Assets.xcassets/favorites illustration.imageset/Contents.json (hunk 1 above)
Comment (finben) [reply Hymns/Assets.xcassets/favorites illustration.imageset/Contents.json]: Lol, my b

File: Hymns/Home/HomeView.swift (hunk 4 above)
Comment (skywalkerdude) [reply Hymns/Home/HomeView.swift]: I'm looking at the way you did it, and it looks like it also pushes up the search bar in landscape. Am I missing something here?
Simulator Screen Shot - iPhone 8 - 2020-05-22 at 17 15 34

File: Hymns/Assets.xcassets/empty favorites illustration.imageset/Contents.json [hunk 6]
```diff
@@ -0,0 +1,21 @@
...
+      "scale" : "1x"
+    },
+    {
+      "idiom" : "universal",
+      "scale" : "2x"
+    },
+    {
+      "idiom" : "universal",
+      "scale" : "3x"
+    }
```
skywalkerdude, line 15: You only provided a file for the 1x version. How does the 2x and 3x work? You don't need to provide images for those also?

Hymns/Home/HomeView.swift
```diff
@@ -23,22 +23,34 @@ struct HomeView: View {
                 .padding(.horizontal)
                 .padding(.top, viewModel.searchActive ? nil : .zero)
 
-            viewModel.label.map {
-                Text($0).fontWeight(.bold).padding(.top).padding(.leading).foregroundColor(Color("darkModeSubtitle"))
+            if !viewModel.songResults.isEmpty {
+                viewModel.label.map {
+                    Text($0).fontWeight(.bold).padding(.top).padding(.leading).foregroundColor(Color("darkModeSubtitle"))
+                }
```
skywalkerdude, line 29: I don't mean to be nitpicky, but I think this is better done in the viewmodel, than the view. It's business logic (as opposed to view logic) and can be unit tested if it lived in the viewmodel instead

skywalkerdude: Since it's technically outside the scope of this PR we can leave a TODO and do this label hiding later as well

Hymns/Home/HomeView.swift
finben: True.... in my new commit, I added the image to be able to resize while maintaining its aspect ratio.

Hymns/Assets.xcassets/empty favorites illustration.imageset/Contents.json
finben: It's wierd. I have tried creating 2x and 3x versions for the 1x and it always looks pixelated and wrong. So then I tried making the initial image the 3x and downgrading by putting in 1x and 2x but it also looks pixelated. It seems to me just by leaving the highest res image we are getting the best result on all devices. I have tried different ways and websites to scale out 2x and 3x but with no luck. To me what we have is looking good on every test emulator so idk...

Hymns/Home/HomeView.swift
finben: Done, just going to leave the TODO.

Hymns/Assets.xcassets/empty favorites illustration.imageset/Contents.json
skywalkerdude: ok, I guess we can try to fix it later if we need to

PR #179, opened 2020-05-24
Change font

skywalkerdude:
Implement the change-font option to the bottom bar

### DIFF
--- a/Hymns/Display/VerseViewModel.swift
+++ b/Hymns/Display/VerseViewModel.swift
@@ -1,5 +1,7 @@
+import Combine
 import Foundation
 import Resolver
+import SwiftUI
 
 class VerseLineViewModel: Hashable, ObservableObject {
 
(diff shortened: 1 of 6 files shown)


Hymns/Display/VerseViewModel.swift
```diff
@@ -1,28 +1,59 @@
...
+                    self.fontSize = fontSize
+                }
+        }.store(in: &disposables)
+    }
+
+    static func == (lhs: VerseLineViewModel, rhs: VerseLineViewModel) -> Bool {
+        lhs.verseNumber == rhs.verseNumber && lhs.verseText == rhs.verseText && lhs.transliteration == rhs.transliteration
+    }
+
+    func hash(into hasher: inout Hasher) {
```
finben, line 35: What does this do? How does the hash function work?

Hymns/Infrastructure/Application/UserDefaultsManager.swift
```diff
@@ -0,0 +1,81 @@
+import Combine
+import Foundation
+import SwiftUI
+
+/**
+ * Wrapper class for managing `UserDefaults`
+ */
+class UserDefaultsManager {
```
finben, line 8: This doesn't have to be an observable object to send values?

Hymns/Display/VerseViewModel.swift
skywalkerdude: Are you asking how a hash function works in general? https://en.wikipedia.org/wiki/Hash_function

Hymns/Infrastructure/Application/UserDefaultsManager.swift
skywalkerdude: It doesn't need to be an observable object because I'm writing it to a CurrentValueSubject that is observed on in the view model

PR #187, opened 2020-05-24
Initial commit. Stores the lyrics for the share.

finben:
Works on devices great. I just need to format the lyrics a tad bit better to have the stanzas separated. The only problem is how to not use the environmentObject.... It is basically the same challenge I was having with the dynamic title change branch. At a certain point it almost just makes sense to use environment object. It becomes such a nightmare to try and pass data between something like VerseView and DisplayHymnBottomToolbar without using an environmentObject. Idk, what you think but in our more complex cases should we start using environmentObject. 

I know from one of the SwiftUI WWDC talks I watched Apple suggested developers could be using EnvironmentObject liberally. Also, most articles I read about it say that once you start having to pass an ObservableObject through to many views you should just consider making it an EnvironmentObject. Obviously, there is a sacrifice to our testing I suppose? What do you think?


skywalkerdude:
Have you tried just fetching it from the repository?

This is where the repository pattern really shines. You don't need to pass anything, because you can get fetch it from the repository.

So in the bottom bar view model, you can just call fetchHymn, and everything you need is right there for you to use.

Hymns/Display/ShareSheet.swift
```diff
@@ -0,0 +1,23 @@
+import SwiftUI
+
+struct ShareSheet: UIViewControllerRepresentable {
```
skywalkerdude, line 3: Could you document what you followed to make this class? Just for posterity

Hymns/Display/ShareSheet.swift
```diff
@@ -0,0 +1,23 @@
...
+    typealias Callback = (_ activityType: UIActivity.ActivityType?, _ completed: Bool, _ returnedItems: [Any]?, _ error: Error?) -> Void
+      
+    let activityItems: [Any]
```
skywalkerdude, line 6: pls fix swiftlint warnings

Hymns/Display/DisplayHymnBottomBar.swift
```diff
@@ -30,10 +32,18 @@ struct DisplayHymnBottomBar: View {
         HStack(spacing: 0) {
             Group {
                 Spacer()
-                Button(action: {}, label: {
-                    BottomBarLabel(imageName: "square.and.arrow.up")
-                })
+                    Button(action: {
+                        self.showShareSheet = true
+                    }) {
```
skywalkerdude, line 37: pls fix swiftlint warnings

finben:
Thanks a lot! This makes so much sense and is great. Just hit the repo. It works like a charm and I like it much better because it is cleaner and easier to understand now what is going on. 

To answer your question about what video it was from https://developer.apple.com/videos/play/wwdc2019/226/   around the 27 minute mark or so.

Hymns/Display/ShareSheet.swift
finben: Done

Hymns/Display/DisplayHymnBottomBar.swift
finben: Done

Hymns/Display/ShareSheet.swift
finben: Added in

skywalkerdude:
loool this article is straight up like, the best way to use env objects: don't use them 😂 

https://medium.com/better-programming/the-best-way-to-use-environment-objects-in-swiftui-d9a88b1e253f

skywalkerdude:
So I read through this: https://swiftwithmajid.com/2019/08/21/the-power-of-environment-in-swiftui/

My biggest gripe with env objects is that it doesn't guarantee any kind of compile-time safety. 

Look at this example
```
struct CalendarView : View {
    @EnvironmentObject var store: CalendarStore

    var body: some View {
        NavigationView {
            List {
                ForEach(self.store.sleeps) { sleep in
                    NavigationLink(
                        destination: SleepDetailsView()
                            .environmentObject(SleepStore(sleep: sleep))
                    ) {
                        CalendarRow(sleep: sleep)
                    }
                }
            }
        }.navigationBarTitle("calendar")
    }
}
```

When you are ini SleepDetailsView(), you have no idea if the environment variable is set by a parent view, and if SleepDetailsView is used ini a different context where the developer forgot to set the SleepStore,  we won't know until we crash at runtime

Hymns/Display/DisplayHymnBottomBar.swift
```diff
@@ -30,16 +31,24 @@ struct DisplayHymnBottomBar: View {
...
+
                 Spacer()
             }
             Group {
                 Button(action: {self.tabPresented = .fontSize}, label: {
                     BottomBarLabel(imageName: "textformat.size")
                 })
                 Spacer()
+            }.onAppear {
+                self.viewModel.fetchLyrics()
```
skywalkerdude, line 51: lol it kind of looks like you just picked a random Group to stick this onAppear{} in.

I would do it on the HStack, just to keep all this stuff in one place. Besides, if for some reason in the future we don't want to show font size based on some param, then this will break. We can pretty much guarantee that the HStack will always be shown, so it's safer to put it there

Hymns/Display/DisplayHymnBottomBar.swift
```diff
@@ -30,16 +31,24 @@ struct DisplayHymnBottomBar: View {
...
-                    BottomBarLabel(imageName: "square.and.arrow.up")
-                })
+                    Button(action: {
+                        self.showShareSheet = true
+                    }, label: {
+                        BottomBarLabel(imageName: "square.and.arrow.up")
+                    })
+                .sheet(isPresented: $showShareSheet) {
+                    ShareSheet(activityItems: [self.viewModel.sharablelLyrics])
+                }
```
skywalkerdude, line 41: I would put this sheet on the HStack, so all that logic is in the same place and better organized.

Also, please use `.sheet(item: Binding<Identifiable?>, content: (Identifiable) -> View)` instead of `.sheet(isPresented: Binding<Bool>, content: () -> View)`

and create a 
```
enum DisplayHymnSheet: String {
    case share = "Share Lyrics"
}
```
and switch on that, kind of like what I did with `DisplayHymnActionSheet`.

The rationale here is that, there could potentially be 7 (or more) different bottom sheets. But there are only 3 **types** of bottom sheets. This reduces the number of @State variables we'll have to manage, by only managing one state variable per **type** of bottom sheet, instead of one per action.

Hymns/Display/DisplayHymnBottomBar.swift
finben: Lol yeah, my bad haha I didn't mean to leave it here. I was just testing something real fast and threw it on here and forgot to move it when it actually ended up working.

Hymns/Display/DisplayHymnBottomBarViewModel.swift
```diff
@@ -5,12 +5,68 @@ import Resolver
...
+
+    func fetchLyrics() {
+        var latestValue: [VerseViewModel]? = [VerseViewModel]()
+        repository
+            .getHymn(identifier)
+            .map({ [weak self] hymn -> [VerseViewModel]? in
+                guard let self = self, let hymn = hymn, !hymn.lyrics.isEmpty else {
+                    return nil
+                }
+                return self.convertToViewModels(verses: hymn.lyrics)
```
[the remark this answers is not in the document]
finben: Good idea, resolved in most recent commit

Hymns/Display/DisplayHymnBottomBar.swift
finben: Nice, makes sense. resolved in most recent commit

finben:
The articles you have are pretty compelling for using it as little as possible. The point is valid. I think the first article his first point is ludicrous since I can't imagine someone would just forget to put the environment object in the scene delegate and be so confused as to not know how to debug that. But the other points are valid and make a strong case against using it at least as little as possible. Thanks.

Hymns/Display/VerseViewModel.swift
```diff
@@ -3,6 +3,9 @@ import Foundation
 import Resolver
 import SwiftUI
 
+/**
+    This class is specifically a way to actually store hymn lyrics for when they need to be shared with other people. The text is simply populated dynamically when the verse views appear and later depopulated when that hymn disappears.
+*/
```
skywalkerdude, line 8: delete pls

Hymns/Display/DisplayHymnBottomBarViewModel.swift
```diff
@@ -5,12 +5,52 @@ import Resolver
...
+                if state == .finished {
+                    // Only display the lyrics if the call is finished and we aren't getting any more values
+                    self.shareableLyrics = latestValue
+                }
+                }, receiveValue: { lyrics in
+                    latestValue = lyrics
+            }).store(in: &disposables)
+    }
+
+    func convertToOneString(verses: [Verse]) -> String {
```
skywalkerdude, line 45: this should be private

Hymns/Display/DisplayHymnBottomBar.swift
```diff
@@ -22,17 +22,26 @@ struct DisplayHymnBottomBar: View {
...
+                }, label: {
                     BottomBarLabel(imageName: "square.and.arrow.up")
-                })
+                    })
+                    .sheet(item: $sheetPresented) { tab -> ShareSheet in
+                        switch tab {
+                        case .share:
+                            return ShareSheet(activityItems: [self.viewModel.shareableLyrics])
+                        }
+                }
```
skywalkerdude, line 44: Could you put this on the HStack instead?

Hymns/Display/DisplayHymnBottomBarViewModel.swift
```diff
@@ -5,12 +5,52 @@ import Resolver
...
+        var latestValue: String = ""
+        repository
+            .getHymn(identifier)
+            .map({ [weak self] hymn -> String in
+                guard let self = self, let hymn = hymn, !hymn.lyrics.isEmpty else {
+                    return ""
+                }
+                return self.convertToOneString(verses: hymn.lyrics)
+            })
+            .receive(on: mainQueue)
```
skywalkerdude, line 33: right above this, could you add a ".subscribe(on: backgroundQueue)"

finben:
@skywalkerdude I added testing but it is passing when it should be failing?... Specifically the last test last lines 

                    it("shareable lyrics should be an empty string") {
                        expect(target.shareableLyrics).to(equal(""))
                    }

It should not be "" and should fail. Do I need to regenerate mock file or something?

skywalkerdude:
which line should not be an empty string? You have three cases of 
```
                it("shareable lyrics should be an empty string") {
                    expect(target.shareableLyrics).to(equal(""))
                }
```

They all look like they should be empty?

HymnsTests/Display/DisplayHymnBottomBarViewModelSpec.swift
```diff
@@ -1,22 +1,85 @@
...
+                        testQueue.sync {}
+                        testQueue.sync {}
+                        testQueue.sync {}
+                    }
+                    it("should call hymnsRepository.getHymn") {
+                        verify(hymnsRepository.getHymn(classic1151)).wasCalled(exactly(1))
+                    }
+                    it("should set up the info dialog") {
+                        expect(target.songInfo).to(equal(SongInfoDialogViewModel(hymnToDisplay: classic1151)))
+                    }
```
skywalkerdude, line 51: You don't really need this because the fetch doesn't affect the song info dialog

HymnsTests/Display/DisplayHymnBottomBarViewModelSpec.swift
```diff
@@ -1,22 +1,85 @@
...
```
skywalkerdude, line 78: You don't really need this because the fetch doesn't affect the song info dialog

HymnsTests/Display/DisplayHymnBottomBarViewModelSpec.swift
```diff
@@ -1,22 +1,85 @@
...
                 }
                 it("should set up the info dialog") {
                     expect(target.songInfo).to(equal(SongInfoDialogViewModel(hymnToDisplay: classic1151)))
                 }
+                it("shareable lyrics should be an empty string") {
+                    expect(target.shareableLyrics).to(equal(""))
+                }
+            }
+
+            context("with valid repository and hymn 1151") {
```
skywalkerdude, line 32: "with nil repository result" since you are returning nil in from the repo

HymnsTests/Display/DisplayHymnBottomBarViewModelSpec.swift
```diff
@@ -1,22 +1,85 @@
...
+                    it("shareable lyrics should be an empty string") {
+                        expect(target.shareableLyrics).to(equal(""))
+                    }
+                }
+            }
+
+            context("fetch hymn 1151 with valid repository") {
+                beforeEach {
+                    target = DisplayHymnBottomBarViewModel(hymnToDisplay: classic1151, hymnsRepository: hymnsRepository, mainQueue: testQueue, backgroundQueue: testQueue)
+                    let hymn = UiHymn(hymnIdentifier: classic1151, title: "title", lyrics: [Verse](), pdfSheet: Hymns.MetaDatum(name: "Lead Sheet", data: [Hymns.Datum(value: "Piano", path: "/en/hymn/c/1151/f=ppdf"), Hymns.Datum(value: "Guitar", path: "/en/hymn/c/1151/f=pdf"), Hymns.Datum(value: "Text", path: "/en/hymn/c/1151/f=gtpdf")]))
```
skywalkerdude, line 61: You just need the `lyrics: [Verse]()` and not all this other stuff since those things are irrelevant in this test

HymnsTests/Display/DisplayHymnBottomBarViewModelSpec.swift
```diff
@@ -1,22 +1,85 @@
...
+                        testQueue.sync {}
+                        testQueue.sync {}
+                    }
+                    it("should call hymnsRepository.getHymn") {
+                        verify(hymnsRepository.getHymn(classic1151)).wasCalled(exactly(1))
+                    }
+                    it("should set up the info dialog") {
+                        expect(target.songInfo).to(equal(SongInfoDialogViewModel(hymnToDisplay: classic1151)))
+                    }
+                    it("shareable lyrics should be an empty string") {
```
skywalkerdude, line 79: Can you also add a case where you return verse text and assert on a non-empty shareable lyrics

finben: Added in most recent commit. I think I am finally getting a better grasp on testing.

HymnsTests/Display/DisplayHymnBottomBarViewModelSpec.swift
finben: Thanks

PR #190, opened 2020-05-24
.primary color is not working in TabView

finben:
I have no idea why foregroundColor(.primary) is not adjusting for dark mode when implemented in TabView. Anyways, just recreated an asset that is the same as .primary and implemented.


skywalkerdude:
I don't see what was wrong with the tab view. It looks fine to me?
Simulator Screen Shot - iPhone 8 - 2020-05-24 at 17 04 40

finben:
Sorry the other TabView. The custom one you see when you are in display hymn

skywalkerdude:
ok. let me check.

Just as an FYI, at work usually when we make UI fixes, we try to include a screenshot (of before and after if applicable), to help make it easier to review. If you are working off an emulator, there's an easy screenshot button at the top of the emulator. Just for next time.

Screen Shot 2020-05-24 at 5 18 30 PM

finben:
Yeah, good point

finben:
Here is a picture of it before the fix in dark mode
Simulator Screen Shot - Clone 1 of iPhone 11 - 2020-05-24 at 21 31 41

skywalkerdude:
> I have no idea why foregroundColor(.primary) is not adjusting for dark mode when implemented in TabView. Anyways, just recreated an asset that is the same as .primary and implemented.

It looks like that because the background is white and the foreground color is also white. The foreground color is correct, while the background is not.

I think you can just use .backgroundColor(.systemBackground)

I fixed it in https://github.com/HymnalDreamTeam/Hymns-iOS/pull/193. take a look and let me know if it works for you.

finben:
Closing and going with the result of #193

Issue #170, opened 2020-05-21
7 Icons Question


finben:
This is regarding the bottom toolbar when you are looking at a hymn. I just wanted to bring this up for consideration. To me the side tolerances look a little tight and 7 icons feels cluttered? Luke, mentioned the other android app also can have up to 7 but most hymns only have data for 6. What is your opinion? To me 6 icons seems to be the max amount that doesn't start feeling like too much. I wonder if we can move something like text size to our settings tab instead of having it right in the toolbar.

Emyeh:
That's a good point, I agree it feels cluttered to have 6+ icons. We originally had all the icons there without considering a hierarchy of value, because we're not 100% sure which features people would use more or less. 

6 is a good amount to leave out for now. I think text size is a good one to put within an ellipses menu. I would assume people only need to adjust once. We have the ellipses for android, but I can also add it to the iOS designs.

skywalkerdude:
@Emyeh I don't think we have ellipses on Android. We just show all 7 tabs (at most) if the hymn has everything. How should we ellipsize on Android?

Emyeh:
I mean I have the ellipses in the android design already. I'll show the full flow for both.

Emyeh:
@finben Added the ellipses menu to iOS and removed the font size icon from the hymn tool bar. Let me know if the custom bottom sheet will take too long, I can try to adjust to the standard iOS bottom sheet - https://www.figma.com/file/7H5ApjzuVuJL3OP7FvE2qe/h-app?node-id=647%3A3155

finben:
@Emyeh Thanks, looks great!! Yeah, I am glad we are doing this. I went to test the app on my wife's small iPhone 7 and was trying to click one of the toolbar tabs and we both kept clicking the other one because they were so close together.

Emyeh:
@finben Awesome! Let me think a little bit more about which features should also be included in the overflow menu. It looks a bit funny now with just the font size option. Will try to have something soon.

skywalkerdude:
I think one of the key questions is: how do we show the overflow menu for the bottom bar?

Since we can have a dynamic number of tabs (anywhere from 3-7), I think the best would be that we prioritize them, and if there are too many, we put the lower priority tabs in an overflow menu.

[...]

PR #192, opened 2020-05-25
Meet new figma spec and move font size option to ellipsis

finben:
Simulator Screen Shot - Clone 1 of iPhone 11 - 2020-05-24 at 21 31 58

Here is a screenshot after the work was done. The functionality of the button remained the exact same.


finben:
Addresses Issue #170

skywalkerdude:
I think emily is rethinking how to do the overflow menu, so maybe we can revisit this in a bit?

PR #199, opened 2020-05-28
Add functionality for going to different languages

skywalkerdude:


skywalkerdude:
hmmm yeah. But SwiftUi doesn't give us a good way to manage the backstack, so it looks like we're stuck with this for now...?

Do you have a better idea?

skywalkerdude:
(created an issue for it https://github.com/HymnalDreamTeam/Hymns-iOS/issues/206)

finben:
Nope :/ we can think of something later. Maybe they will give us some new tools at WWDC 2020 in a few weeks

PR #209, opened 2020-05-29
Initial EarlGrey 1 UI testing

finben:
Maybe we can do everything we need to with version 1.0 of EarlGrey. It seems more stable at the moment. I also, had trouble getting EarlGrey 2.0 setup. Here is an initial UI Test. It takes advantage of the asynchronous feature of EarlGrey automatically and waits for the search to complete without having to put in wait or sleep. We should be able to set it up with dependency injection too. I need to look into that more tomorrow. 

Here is a good source for learning EarlGrey. Probably the best I have found so far. 
https://www.youtube.com/watch?v=CEekHqcnPKQ


HymnsUITests/EarlGrey.swift
```diff
@@ -0,0 +1,188 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import EarlGrey
+import Foundation
+
+public func GREYAssert(_ expression: @autoclosure () -> Bool, reason: String) {
+  GREYAssert(expression(), reason, details: "Expected expression to be true")
+}
```
skywalkerdude, line 22: What are these things? Are these part of the set up? Or are these tests? I can't tell

skywalkerdude:
`Please note that EarlGrey 2.0 uses a UI Testing Target and not a Unit Testing Target like EarlGrey 1.0.`

Do you know what ramifications this would have? Where we are using a unit testing target too do ui testing?

HymnsUITests/EarlGrey.swift
finben: EarlGrey.swift is an autogenerated file as part of the setup to give us these functions for the tests. EarlGreyExtensions.swift are also extensions to give us more functionality for use in our tests. Only EarlGreyUITests.swift are the tests that I wrote.

finben:
The only real ramification I can think of is that it could slow down our unit test suite a bit because earl grey tests will need to run. There are also these advantage points from the EarlGrey 2.0 site I copied below. But I'm not sure how much they really apply or will impact us personally...

EarlGrey 2.0 advantages over EarlGrey 1.0
- Out of Process Testing using XCUITest. So System Alerts, Inter-app interactions etc. are supported
- Lesser throttling of the application under test's main thread.

finben:
One other annoyance is getting the state to reset after each test. It is in their FAQ but they don't show you how to actually write the resetAppDelegate() method.... https://github.com/google/EarlGrey/blob/master/docs/faq.md  As of now I haven't been able to get the application to reset after each EarlGrey test

Hymns.xcworkspace/xcshareddata/WorkspaceSettings.xcsettings
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>PreviewsEnabled</key>
+	<false/>
+</dict>
+</plist>
```
skywalkerdude, line 8: whats this?

HymnsTests/Earl Grey Tests/EarlGreyUITests.swift
```diff
@@ -0,0 +1,105 @@
+import XCTest
+import EarlGrey
+@testable import Hymns
+
+class EarlGreyUITests: XCTestCase {
+
+    override func setUp() {
```
skywalkerdude, line 7: Can you make a new target for these? So that way we have some separation between unit tests and ui test?

HymnsTests/Earl Grey Tests/EarlGreyUITests.swift
```diff
@@ -0,0 +1,105 @@
...
+        EarlGrey.selectElement(with: grey_accessibilityLabel("chevron.left"))
+            .assert(grey_sufficientlyVisible())
+            .perform(grey_tap())
+        EarlGrey.selectElement(with: grey_accessibilityLabel("Cancel"))
+            .assert(grey_sufficientlyVisible())
+            .perform(grey_tap())
+    }
+}
+
+//TODO: Find a way to reset the app after each test as mentioned at https://github.com/google/EarlGrey/blob/master/docs/faq.md then we can combine all the tests in one class.
```
skywalkerdude, line 44: Do we want to combine all the tests into one file? Wouldn't that make things really difficult to debug if something goes wrong?

finben:
Migrating to #213

Hymns.xcworkspace/xcshareddata/WorkspaceSettings.xcsettings
finben: This is an auto generated plist file by earl grey. generated through installing the pods

HymnsTests/Earl Grey Tests/EarlGreyUITests.swift
finben: Yeah, good idea. I am calling it functionalTesting because I noticed that is what a lot of other people call this kind of thing like KIF which is a similar type of UI testing using the unit testing from Xcode

HymnsTests/Earl Grey Tests/EarlGreyUITests.swift
finben: I don't think so because the test navigation will always show exactly which class failed the test? Not just the file that failed. If you want I can move them to separate files?

PR #210, opened 2020-05-29
Add initial code to display the browse categories

skywalkerdude:
Add initial code to display the browse categories. Unable to test on device/simulator yet, can only test with live preview


Hymns/Browse/BrowseCategoriesViewModel.swift
```diff
@@ -0,0 +1,51 @@
+import Combine
+import Foundation
```
finben, line 2: I don't think you need Foundation here

Hymns/Browse/CategoryViewModel.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation
```
finben, line 1: I don't think you need Foundation here

Hymns/Repository/CategoriesRepository.swift
```diff
@@ -0,0 +1,27 @@
+import Combine
+import Foundation
```
finben, line 2: I don't think you need Foundation here

Hymns/Browse/CategoryViewModel.swift
skywalkerdude: we need it for `ObservableObject`

PR #215, opened 2020-05-30
build on top of existing browse work to bring browse categories into the browse view

skywalkerdude:


finben:
Ideally, we wouldn't use a fixed number for the frame. Maybe a geometry reader could be used here. But basically your subcategory list here needs a frame since it is a nested list. The view needs to know how much space to alot for the second list. 

answer number 2
https://stackoverflow.com/questions/58882673/swiftui-nested-list

skywalkerdude:
nice, thanks!!

PR #221, opened 2020-05-30
Lottie launch

finben:
I propose we just get this in for now with it only showing animation on the first time you install. And then we add as an improvement issue to have it also display if we are still pinging the API for information or need more time for loading. 

I wasn't able to find a way to actually have the animation run through the launchstoryboard besides static images. Primarily, it doesn't seem to let you hook it up programmatically which is what I would need to connect the Lottie viewController stuff. This guy from Reddit suggests why this may be the case. 

"You cannot create a launch screen programmatically because that defeats the entire purpose of having a launch screen. The whole point of a launch screen is to show you something static that can be rendered as quickly as possible while your application is launching. If iOS waited to load your application so that it could execute your code, the point at which the launch screen is useful will have passed." 

https://www.reddit.com/r/iOSProgramming/comments/b4kyed/programmatic_launchscreen/


skywalkerdude:
Hmm, okay. If you can't change LaunchScreen.storyboard programmatically, then let's just stick the logo into LaunchScreen.storyboard and keep it at that. Creating a second animation afterwards doesn't do anything because now the user is going to see two "splash" screens.

skywalkerdude:
I'm really confused what this is saying: https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/launch-screen/

It says it should be close to the first screen, but also that it shouldn't have text... what if your first screen has text -__-

I feel like most apps on the app store doesn't follow this guidance lol

finben:
@skywalkerdude yeah, that is weird. I honestly wonder if a bunch of apps aren't doing what I did here which is it is really a pseudo launch screen. Like the screen has launched already and then they do the launch. Look at the raywenderlich link I sent you. They use a color screen as the launch. Let me send you a Facebook video...

skywalkerdude:
Looks like something in this PR is causing `HymnDataStoreGrdbImplSpec` to crash with 
```
*** Terminating app due to uncaught exception 'RLMException', reason: 'Realm accessed from incorrect thread.'
terminating with uncaught exception of type NSException
abort() called
CoreSimulator 704.12.2 - Device: Clone 1 of iPhone 11 Pro (CFE5EBAB-56E9-44DF-AF2B-A05DAAF5A851) - Runtime: iOS 13.4 (17E8260) - DeviceType: iPhone 11 Pro
```

which is super bizarre given that HymnDataStoreGrdbImplSpec has nothing to do with realm. I can't figure out why it's happening. Do you have any idea why that would be happening?

skywalkerdude:
As a last resort, you can create a new PR, and make each commit step by step and run the test plan to see what introduced the error. It's tedious (hence, a last resort), but I'm fairly confident that it's introduced by this PR, since it passes in all the other PRs and at master.

finben:
Migrated to #274

Issue #186, opened 2020-05-24
Add hymn title to classic hymn numbers in the search result


skywalkerdude:
![100656211_621453128581617_4236500168192557056_n](https://user-images.githubusercontent.com/1427524/82746982-10031680-9d4a-11ea-9781-49f91946fac6.jpg)
Currently we strip out the number for classic hymns.

But it'd be clearer if it showed `Hymn 82 - Down from His glory` instead of just `Hymn 82`

skywalkerdude:
lol... @rachely3n I think you actually added it to the wrong place >.>

I meant to add it to the search results, not to the song itself. So the work should actually be done in `HomeViewModel` and it should replace `SongResultViewModel(title: recentSong.songTitle` with the correct title... >.< my bad for giving you the wrong guidance

PR #222, opened 2020-05-31
Changing classic hymn titles to include first line of verse

rachely3n:
Addressing #186. I will squash my commits before merging.

Note that the title ends in punctuation since a lot of the lyric lines end in some sort of punctuation. Is that alright or should we strip it?
Screen Shot 2020-05-30 at 6 23 34 PM

### DIFF
--- a/Hymns/Display/DisplayHymnViewModel.swift
+++ b/Hymns/Display/DisplayHymnViewModel.swift
@@ -64,13 +64,21 @@ class DisplayHymnViewModel: ObservableObject {
                     guard let self = self else { return }
                     guard let hymn = hymn else { return }
 
+                    //NOTE: 224, it turns into As we're sharing of the cup,
                     let title: Title
                     if self.identifier.hymnType == .classic {
-                        title = "Hymn \(self.identifier.hymnNumber)"
+                        var titleLyric = ""
+                        if let lyrics = hymn.lyrics.first {
+                            if let firstLine = lyrics.verseContent.first {
+                                titleLyric = ": \(firstLine)"
+                            }
+                        }
+                        title = "Hymn \(self.identifier.hymnNumber)\(titleLyric)"
                     } else {
                         title = hymn.title.replacingOccurrences(of: "Hymn: ", with: "")
                     }
                     self.title = title
+                    print("meep: \(title)")
 
                     self.tabItems = [self.currentTab]
 
--- a/Hymns/Display/DisplayHymnViewModel.swift
+++ b/Hymns/Display/DisplayHymnViewModel.swift
@@ -64,18 +64,13 @@ class DisplayHymnViewModel: ObservableObject {
                     guard let self = self else { return }
                     guard let hymn = hymn else { return }
 
-                    let title: Title
+                    var title: Title
                     if self.identifier.hymnType == .classic {
-                        var titleLyric = ""
-                        if let lyrics = hymn.lyrics.first {
-                            if let firstLine = lyrics.verseContent.first {
-                                titleLyric = ": \(firstLine)"
-                            }
-                        }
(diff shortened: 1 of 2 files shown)


rachely3n:
Don't understand why test passes for me locally but fails here...

skywalkerdude:
they're flaky :(

Hymns/Display/DisplayHymnViewModel.swift
```diff
@@ -66,7 +66,13 @@ class DisplayHymnViewModel: ObservableObject {
 
                     let title: Title
                     if self.identifier.hymnType == .classic {
-                        title = "Hymn \(self.identifier.hymnNumber)"
+                        var titleLyric = ""
+                        if let lyrics = hymn.lyrics.first {
+                            if let firstLine = lyrics.verseContent.first {
+                                titleLyric = ": \(firstLine)"
+                            }
+                        }
```
skywalkerdude, line 74: I don't think you need to do this. you have access to hymn.title, and you can just use that.

So it'd be like
```
let title = ""
if hymnType == classic {
  title = "Hymn \(hymnumber): "
}
title.append(hymn.title.replacingOccurrences(of: "Hymn: ", with: ""))
```

HymnsTests/Display/DisplayHymnViewModelSpec.swift
```diff
@@ -54,7 +54,8 @@ class DisplayHymnViewModelSpec: QuickSpec {
                         beforeEach {
                             target = DisplayHymnViewModel(backgroundQueue: testQueue, favoritesStore: favoritesStore, hymnToDisplay: classic1151, hymnsRepository: hymnsRepository, historyStore: historyStore,
                                                           mainQueue: testQueue, pdfPreloader: pdfLoader, storeInHistoryStore: true)
-                            let hymn = UiHymn(hymnIdentifier: classic1151, title: "title", lyrics: [Verse](), pdfSheet: Hymns.MetaDatum(name: "Lead Sheet", data: [Hymns.Datum(value: "Piano", path: "/en/hymn/c/1151/f=ppdf"), Hymns.Datum(value: "Guitar", path: "/en/hymn/c/1151/f=pdf"), Hymns.Datum(value: "Text", path: "/en/hymn/c/1151/f=gtpdf")]))
+                            let hymn = UiHymn(hymnIdentifier: classic1151, title: "title", lyrics: [Verse(verseType: VerseType.verse, verseContent: ["Drink! A river pure and clear that’s flowing from the throne;"])], pdfSheet: Hymns.MetaDatum(name: "Lead Sheet", data: [Hymns.Datum(value: "Piano", path: "/en/hymn/c/1151/f=ppdf"), Hymns.Datum(value: "Guitar", path: "/en/hymn/c/1151/f=pdf"), Hymns.Datum(value: "Text", path: "/en/hymn/c/1151/f=gtpdf")]))
```
skywalkerdude, line 57: Then you can revert this and just use the fake `title` instead

Hymns/Display/DisplayHymnViewModel.swift
skywalkerdude: I think that also takes care of your punctuation problem!

Hymns/Display/DisplayHymnViewModel.swift
```diff
@@ -64,12 +64,13 @@ class DisplayHymnViewModel: ObservableObject {
                     guard let self = self else { return }
                     guard let hymn = hymn else { return }
 
-                    let title: Title
+                    var title: Title
```
skywalkerdude, line 67: I think this can still be a let

rachely3n: It complained later when I did the append since let made it immutable.

PR #225, opened 2020-05-31
Lottie

finben:


finben:
What do you want me to do about the danger notification about copywriter headers? Don't we need to leave them in the files since they are not our pods? Or can't we just make Danger + SwiftLint not apply to our Pod files?

finben:
Nothing I have done will cause Lottie to pass tests. Lottie is the problem. I am giving up and going with just the static launch like you mentioned. Let's not delete this branch though... In case we ever decide to try and implement this down the road.

PR #226, opened 2020-05-31
Gave up on Lottie. Going with static image on launch only.

finben:
I could never get Lottie to pass tests. Got frustrated and giving up. Maybe we can keep the Lottie branch around in case we ever decide to implement it later.


skywalkerdude:
This is what it looks like in dark mode:
Simulator Screen Shot - iPhone 8 - 2020-05-31 at 16 49 56

Could you export the static image with the transparent background instead of white background?

finben:
I couldn't figure out a good way to get a transparent background based off what Emily gave me. But she did give me dark mode images with the dark background so I just put that in instead.

PR #229, opened 2020-06-02
Core data is actually pretty easy to use... o.O

finben:
Main sources are 
https://www.youtube.com/watch?v=7_Afen3PlDE
https://www.youtube.com/watch?v=cuHzVt0VXJM

### DIFF
--- a/Hymns.xcodeproj/project.pbxproj
+++ b/Hymns.xcodeproj/project.pbxproj
@@ -133,6 +133,9 @@
 		5CA7849E242D14C800B57903 /* FavoritesView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5CA7849D242D14C800B57903 /* FavoritesView.swift */; };
 		5CA784A0242D1A3200B57903 /* BrowseView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5CA7849F242D1A3200B57903 /* BrowseView.swift */; };
 		5CA784A4242D1CF800B57903 /* HomeView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5CA784A3242D1CF800B57903 /* HomeView.swift */; };
+		B5032AD124859F99002F6C19 /* TagSheetView.swift in Sources */ = {isa = PBXBuildFile; fileRef = B5032AD024859F99002F6C19 /* TagSheetView.swift */; };
+		B5032AD32485AC3A002F6C19 /* TagListView.swift in Sources */ = {isa = PBXBuildFile; fileRef = B5032AD22485AC3A002F6C19 /* TagListView.swift */; };
+		B5032AD52485AF62002F6C19 /* TagSubSelectionList.swift in Sources */ = {isa = PBXBuildFile; fileRef = B5032AD42485AF62002F6C19 /* TagSubSelectionList.swift */; };
 		B51BA5C92430075300CA52F7 /* DisplayHymnView.swift in Sources */ = {isa = PBXBuildFile; fileRef = B51BA5C82430075300CA52F7 /* DisplayHymnView.swift */; };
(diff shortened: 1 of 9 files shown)


skywalkerdude:
Are you still migrating everything to use Core Data? Are there significant benefits of Core Data vs Realm? I thought we looked at the videos and articles and chose Realm because it just becomes a file that we can inspect with the Realm inspector and that we can easily export and import if we need to do that. How come we are walking back that decision?

finben:
@skywalkerdude I am kind of torn... Today, I started both actually lol. I tried to move everything to Realm first and when I got stuck there. I went to move everything to Core Data. 

I am not totally sure which is better. The benefit to core data is we aren't relying on a 3rd party library and I ended up actually finding it really easy to use. The Realm inspector is of no use anymore since we only got a 30 day free trial with it. So we can't even use that anymore. Realm may have some other benefits though. The amount of data we are going to be storing with either one is actually very small so I don't think it will matter to much in terms of performance. 

Whatever, you think is fine. I am fine with either. Just let me know what you think.

skywalkerdude:
what did you get stuck with for realm?

finben:
I was trying to register a TagStore and Xcode kept saying it could not cast it from TagStoreRealmImpl to TagStore protocol. But the other thing is I am not sure how the data will be structured. Right now we have a primary key on hymnIdentifier and we don't even store a field for favorite. We just add and delete objects and if they are in the db then they are favorites. With tags I am not sure the best way to do this. It's a many to many relationship I guess between tags and hymns. But then do I also treat favorites as its own field now?

skywalkerdude:
> I was trying to register a TagStore and Xcode kept saying it could not cast it from TagStoreRealmImpl to TagStore protocol.
This sounds like a swift protocol problem, not a realm problem, right?

>  But the other thing is I am not sure how the data will be structured.
Here is the schema for the tags on Android:
```
{
  "formatVersion": 1,
  "database": {
    "version": 3,
    "identityHash": "72b84ee868ece7273f2d6a3a1e700be5",
    "entities": [
      {
        "tableName": "TAGS",
        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`HYMN_TYPE` TEXT NOT NULL, `HYMN_NUMBER` TEXT NOT NULL, `QUERY_PARAMS` TEXT NOT NULL, `TAG` TEXT NOT NULL, `SONG_TITLE` TEXT, PRIMARY KEY(`HYMN_TYPE`, `HYMN_NUMBER`, `QUERY_PARAMS`, `TAG`))",
        "fields": [
          {
            "fieldPath": "hymnType",
            "columnName": "HYMN_TYPE",
            "affinity": "TEXT",
            "notNull": true
          },
          {
            "fieldPath": "hymnNumber",
            "columnName": "HYMN_NUMBER",
            "affinity": "TEXT",
            "notNull": true
          },
          {
            "fieldPath": "queryParams",
            "columnName": "QUERY_PARAMS",
            "affinity": "TEXT",
            "notNull": true
          },
          {
            "fieldPath": "tag",
            "columnName": "TAG",
            "affinity": "TEXT",
            "notNull": true
          },
          {
            "fieldPath": "songTitle",
            "columnName": "SONG_TITLE",
            "affinity": "TEXT",
            "notNull": false
          }
        ],
        "primaryKey": {
          "columnNames": [
            "HYMN_TYPE",
            "HYMN_NUMBER",
            "QUERY_PARAMS",
            "TAG"
          ],
          "autoGenerate": false
        },
        "indices": [],
        "foreignKeys": []
      }
    ]
  }
}
```
Basically the schema is 5 fields: hymn type, hymn number, query params, tag, title with the first 4 fields as primary keys.

finben:
Hmmm, yeah that makes sense. Can Favorites just be its own attribute as a Boolean? Seems like it would be easier that way?

skywalkerdude:
I think that just takes up more space without giving much of a speed boost. Cuz now you have an extra column that every entry will need to have a value for. Writes will be slower because you have to write another field, and reads won't be faster because you still need to go through every entry anyway to find the entries with the favorites bit as true

PR #230, opened 2020-06-02
Implement going to the song results for each category

skywalkerdude:


Hymns/Browse/BrowseResultsListView.swift
```diff
@@ -0,0 +1,41 @@
+import SwiftUI
+import UIKit
```
finben, line 2: Why do you need UIKit?

skywalkerdude: removed!

Hymns/Browse/BrowseResultsListView.swift
```diff
@@ -0,0 +1,41 @@
...
+
+struct BrowseResultsListView: View {
+
+    @Environment(\.presentationMode) var presentationMode
+    let viewModel: BrowseResultsListViewModel
```
finben, line 6: Make this a private let and add initializer as we normally do? I have already done this and the UIKit and can just push them to your branch if you agree?

skywalkerdude: Sure, sounds good!. I already removed the UiKit, but you can push the private let + initializer

PR #232, opened 2020-06-02
Increase touch target on back and favorite button in the display hymn screen

skywalkerdude:


finben:
Fyi, I found out that this is somewhat a bug in SwiftUI that should be fixed in their next version. The tappable space is too small on alot of buttons.

skywalkerdude:
oh interesting. What should we do then? Should we just leave it? We can just push this in, and once they fix it, then revert it, so in the meantime it'll be easier to use?

PR #233, opened 2020-06-03
Realm Favorite Tag

finben:
Take a look. I am going to try and keep it broken up into smaller PR's for you. This PR will only function to move our existing functionality of favorites into a tags string.

### DIFF
--- a/Hymns/Display/DisplayHymnViewModel.swift
+++ b/Hymns/Display/DisplayHymnViewModel.swift
@@ -122,7 +122,7 @@ class DisplayHymnViewModel: ObservableObject {
                     }
             }).store(in: &disposables)
     }
-    
+
     func fetchFavoriteStatus() {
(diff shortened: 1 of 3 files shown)


finben:
Wtheck, these are the same tests failing as when I had added Lottie.... So wierd.

finben:
Hey, so as I was doing this I had a question. If tags is part of the primary key then I really can't query primary key? Because I be able to provide all the tags a hymn may have for the query? 

As an example function isFavorite. How would I be able to query this on primary key if I don't know all of the tags up front.
`      guard let queriedObject = realm.object(ofType: FavoriteEntity.self, forPrimaryKey: FavoriteEntity.createPrimaryKey(hymnIdentifier: hymnIdentifier)`

skywalkerdude:
Can you use the filtering function: https://realm.io/docs/swift/latest/#filtering?

finben:
Moving this to realmTags3 PR and closing this one.

PR #235, opened 2020-06-03
Realm tags for favorites and tags.

finben:
Sorry, its a big PR lol... I didn't do a good job of breaking it up. Anyways, I think there is definitely some things that need to be done to clean this up like combine some of the functions if possible etc. But take a look and sanity check it. 

I also, plan to add tests after the first round of PR review.

### DIFF
--- a/Hymns/Display/DisplayHymnBottomBar.swift
+++ b/Hymns/Display/DisplayHymnBottomBar.swift
@@ -20,7 +20,6 @@ struct BottomBarLabel_Previews: PreviewProvider {
 
 struct DisplayHymnBottomBar: View {
 
-    //TODO: REalm variables
     @Binding var dialogBuilder: (() -> AnyView)?
     @State private var actionSheet: ActionSheetItem?
     @State private var sheet: DisplayHymnSheet?
--- a/Hymns/Display/DisplayHymnBottomBar.swift
+++ b/Hymns/Display/DisplayHymnBottomBar.swift
@@ -20,6 +20,7 @@ struct BottomBarLabel_Previews: PreviewProvider {
 
 struct DisplayHymnBottomBar: View {
 
+    //TODO: REalm variables
     @Binding var dialogBuilder: (() -> AnyView)?
     @State private var actionSheet: ActionSheetItem?
     @State private var sheet: DisplayHymnSheet?
@@ -84,7 +85,9 @@ struct DisplayHymnBottomBar: View {
                     }
                 }
                 Group {
-                    Button(action: {}, label: {
+                    Button(action: {
+                        self.sheet = .tags
+                    }, label: {
                         BottomBarLabel(imageName: "tag")
                     })
                     Spacer()
@@ -170,10 +173,12 @@ struct DisplayHymnBottomBar: View {
                                 })
                             }) + [.cancel()])
                 }
-            }.sheet(item: $sheet) { tab -> ShareSheet in
+            }.sheet(item: $sheet) { tab -> AnyView in
                 switch tab {
                 case .share:
-                    return ShareSheet(activityItems: [self.viewModel.shareableLyrics])
+                    return ShareSheet(activityItems: [self.viewModel.shareableLyrics]).eraseToAnyView()
+                case .tags:
+                    return TagSheetView(title: self.viewModel.title, hymnIdentifier: self.viewModel.identifier).eraseToAnyView()
                 }
             }
         }
@@ -194,6 +199,7 @@ extension ActionSheetItem: Identifiable {
 
 enum DisplayHymnSheet: String {
     case share = "Share Lyrics"
+    case tags = "Tags"
 }
 
 extension DisplayHymnSheet: Identifiable {
--- a/Hymns/Display/DisplayHymnBottomBar.swift
+++ b/Hymns/Display/DisplayHymnBottomBar.swift
@@ -177,7 +177,7 @@ struct DisplayHymnBottomBar: View {
                 case .share:
                     return ShareSheet(activityItems: [self.viewModel.shareableLyrics]).eraseToAnyView()
                 case .tags:
-                    return TagSheetView(title: self.viewModel.title, hymnIdentifier: self.viewModel.identifier).eraseToAnyView()
+                    return TagSheetView(viewModel: TagSheetViewModel(hymnToDisplay: self.viewModel.identifier)).eraseToAnyView()
                 }
             }
         }
--- a/Hymns/Display/DisplayHymnBottomBarViewModel.swift
+++ b/Hymns/Display/DisplayHymnBottomBarViewModel.swift
@@ -4,13 +4,11 @@ import Resolver
 
 class DisplayHymnBottomBarViewModel: ObservableObject {
 
-    typealias Title = String
     @Published var songInfo: SongInfoDialogViewModel
     @Published var shareableLyrics: String = ""
     @Published var languages = [SongResultViewModel]()
     @Published var relevant = [SongResultViewModel]()
     @Published var mp3Path: URL?
-    @Published var title: String = ""
 
     private let analytics: AnalyticsLogger
     let identifier: HymnIdentifier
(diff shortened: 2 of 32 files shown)


finben:
Fyi, I started changing the names from FavoritesViewModel to TagViewModel but I realized if I go forward making a bunch of name changes right now that will make the PR even more confusing than it already is. So I will wait for your review to finish name changing things like the mocks etc.

Hymns/Tags/TagSheetView.swift
```diff
@@ -0,0 +1,71 @@
...
+                    Spacer()
+                    Button(action: {
+                        self.presentationMode.wrappedValue.dismiss()
+                    }, label: {
+                        Image(systemName: "xmark.circle")
+                    })
+                }
+                HStack {
+                    TextField("Add tags", text: self.$tagNames)
+                    Button("Add") {
```
skywalkerdude, line 34: Check to make sure you can't add empty tags

Hymns/Favorites/FavoritesView.swift
```diff
@@ -4,30 +4,30 @@ import SwiftUI
 
 struct FavoritesView: View {
 
-    @ObservedObject private var viewModel: FavoritesViewModel
+    @ObservedObject private var viewModel: TagsViewModel
```
skywalkerdude, line 7: This shouldn't be renamed. A view model should be associated with the view that it's supporting, so just because under the hood, Favorites the same thing as tags, doesn't mean we should expose that to the view. That should be abstracted away from the view.

Hymns/Tags/TagSheetView.swift
```diff
@@ -0,0 +1,71 @@
...
+                    }
+                }
+            }
+            Section {
+                HStack {
+                    ForEach(self.viewModel.tags, id: \.self) { tag in
+                        Button(action: {
+                            //TODO FIX Delete is deleting all the tags because the button when clicked is clicking all of them.
+                            self.favoritesStore.deleteTag(primaryKey: FavoriteEntity.createPrimaryKey(hymnIdentifier: self.hymnIdentifier, tags: tag.title), tags: tag.title)
+                            self.viewModel.fetchTagsByHymn(hymnSelected: self.hymnIdentifier)
```
skywalkerdude, line 46: Stuff like this should be in the view model

Hymns/Display/DisplayHymnBottomBar.swift
```diff
@@ -170,10 +172,12 @@ struct DisplayHymnBottomBar: View {
...
                             }) + [.cancel()])
                 }
-            }.sheet(item: $sheet) { tab -> ShareSheet in
+            }.sheet(item: $sheet) { tab -> AnyView in
                 switch tab {
                 case .share:
-                    return ShareSheet(activityItems: [self.viewModel.shareableLyrics])
+                    return ShareSheet(activityItems: [self.viewModel.shareableLyrics]).eraseToAnyView()
+                case .tags:
+                    return TagSheetView(title: self.viewModel.title, hymnIdentifier: self.viewModel.identifier).eraseToAnyView()
```
skywalkerdude, line 180: I'd prefer if you just pass in the identifier and use the Repository to get the title. We don't want to create the idea that we can start passing params that we can fetch from the repo

Hymns/Display/DisplayHymnBottomBarViewModel.swift
```diff
@@ -42,6 +44,14 @@ class DisplayHymnBottomBarViewModel: ObservableObject {
...
                     }
 
+                    let title: Title
+                    if self.identifier.hymnType == .classic {
+                        title = "Hymn \(self.identifier.hymnNumber)"
+                    } else {
+                        title = hymn.title.replacingOccurrences(of: "Hymn: ", with: "")
+                    }
+                    self.title = title
+
```
skywalkerdude, line 54: Don't think you'll need any of this once you remove the code where you pass in the title

Hymns/Favorites/FavoritedEntity.swift
```diff
@@ -5,30 +5,33 @@ class FavoriteEntity: Object, Identifiable {
     @objc dynamic var primaryKey: String!
```
skywalkerdude, line 5: Like I said above, you should create a TagEntity, and have FavoriteEntity subclass it with a special hardcoded string as the tag. So both entities should exist, and we can still use FavoriteEntity as we use it today

Hymns/Display/DisplayHymnViewModel.swift
```diff
@@ -129,12 +129,13 @@ class DisplayHymnViewModel: ObservableObject {
...
     }
+
     func toggleFavorited() {
         isFavorited.map { isFavorited in
             if isFavorited {
-                favoritesStore.deleteFavorite(primaryKey: FavoriteEntity.createPrimaryKey(hymnIdentifier: self.identifier))
+                favoritesStore.deleteTag(primaryKey: FavoriteEntity.createPrimaryKey(hymnIdentifier: self.identifier, tags: "favorited"), tags: "favorited")
             } else {
-                favoritesStore.storeFavorite(FavoriteEntity(hymnIdentifier: self.identifier, songTitle: self.title))
+                favoritesStore.storeFavorite(FavoriteEntity(hymnIdentifier: self.identifier, songTitle: self.title, tags: "favorited"))
```
skywalkerdude, line 138: The hardcoded tag should be a little more unique. It's feasible someone might tag their song as "favorited," and it'll suddenly be a Favorite, instead of a song tagged as "favorited." Throw some underscores or something in there to make it less likely someone would pick that as a tag.

finben: Addressed in commit 6fd7962 to give it the string "_*_favorited_*_"

Hymns/Display/DisplayHymnBottomBarViewModel.swift
finben: Removed in 783aae0

Hymns/Tags/TagSheetView.swift
```diff
@@ -0,0 +1,71 @@
...
+                        self.favoritesStore.storeFavorite(FavoriteEntity(hymnIdentifier: self.hymnIdentifier, songTitle: self.title, tags: self.tagNames))
+                        self.viewModel.fetchTagsByHymn(hymnSelected: self.hymnIdentifier)
+                    }
+                }
+            }
+            Section {
+                HStack {
+                    ForEach(self.viewModel.tags, id: \.self) { tag in
+                        Button(action: {
+                            //TODO FIX Delete is deleting all the tags because the button when clicked is clicking all of them.
```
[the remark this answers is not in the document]
finben: Unfortunately, not. It's like a UI facing bug with the Buttons being in an HStack. Idk, maybe its even the simulator. But I will wait until we get our specs from Emily to address since that may change the way the UI here is structured

Hymns/Favorites/FavoritesView.swift
finben: Addressed in 783aae0

Hymns/Tags/TagSheetView.swift
finben: Addressed in 783aae0

Hymns/Display/DisplayHymnBottomBar.swift
finben: Addressed in 783aae0

skywalkerdude:
Ok, I'm just gonna create a new PR with incremental changes.

Because everything here is so intertwined, I can't just explain one thing without breaking other things. Also things are pretty confusing because there are some things called Tags and some things called Favorites that are really tags.

Can you just cherrypick your changes on top of https://github.com/HymnalDreamTeam/Hymns-iOS/pull/239? Then we can go off that one instead.

Hymns/Favorites/FavoritedEntity.swift
skywalkerdude: here, it's not about abstraction. It's about being DRY. You don't want to have to repeat your hardcoded string everywhere as you do now, because that is brittle and easy to break. I basically implemented what I was saying here: https://github.com/HymnalDreamTeam/Hymns-iOS/pull/239. Let me know if that makes sense to you.

finben:
Closing this branch down. Everything from here I want has either been implemented or moved to #266

PR #240, opened 2020-06-04
Update TagStore with all the methods needed for tags

finben:
Okay, so do you want to review this logic, merge into tag, and then I do the next piece with another PR etc and keep merging into tag... Or should I start committing each piece onto this PR and merge it all as one into tag?


Hymns/Tags/TagStore.swift
```diff
@@ -60,6 +75,21 @@ class TagStoreRealmImpl: TagStore {
             action(favorite)
         }.toNotification()
     }
+
+    // MARK: Realm Query Functions
+
```
skywalkerdude, line 80: What is this comment for?

Hymns/Tags/TagStore.swift
```diff
@@ -60,6 +75,21 @@ class TagStoreRealmImpl: TagStore {
...
+    /** Can be used either with a value to specificially query for one tag or without the optional to query all tags besides favorites*/
```
skywalkerdude, line 81: what do you mean "besides favorites"

skywalkerdude: ohhh I get it

Hymns/Tags/TagStore.swift
```diff
@@ -60,6 +75,21 @@ class TagStoreRealmImpl: TagStore {
...
+    func querySelectedTags(tagSelected: String?) -> Results<TagEntity> {
+        guard let specificTag = tagSelected else {
+            return realm.objects(TagEntity.self)
+        }
+        return realm.objects(TagEntity.self).filter(NSPredicate(format: "tags == %@", specificTag))
+    }
+
+    func queryTagsForHymn(hymnIdentifier: HymnIdentifier) -> Results<TagEntity> {
+        let convertedKey = TagEntity.createPrimaryKey(hymnIdentifier: hymnIdentifier, tag: "")
+        return realm.objects(TagEntity.self).filter(NSPredicate(format: "primaryKey CONTAINS[c] %@ AND tags != %@", convertedKey, "_*_favorited_*_"))
```
skywalkerdude, line 91: Use the static constant in the FavoritesEntity instead of hardcoded string

Hymns/Tags/TagStore.swift
```diff
@@ -60,6 +75,21 @@ class TagStoreRealmImpl: TagStore {
...
+
+    // MARK: Realm Query Functions
+
+    /** Can be used either with a value to specificially query for one tag or without the optional to query all tags besides favorites*/
+    func querySelectedTags(tagSelected: String?) -> Results<TagEntity> {
+        guard let specificTag = tagSelected else {
+            return realm.objects(TagEntity.self)
+        }
+        return realm.objects(TagEntity.self).filter(NSPredicate(format: "tags == %@", specificTag))
+    }
```
skywalkerdude, line 87: Does this really exclude the Favorites tags?

Hymns/Tags/TagStore.swift
```diff
@@ -60,6 +75,21 @@ class TagStoreRealmImpl: TagStore {
...
+
+    func queryTagsForHymn(hymnIdentifier: HymnIdentifier) -> Results<TagEntity> {
+        let convertedKey = TagEntity.createPrimaryKey(hymnIdentifier: hymnIdentifier, tag: "")
```
skywalkerdude, line 90: Do you need this? I feel like you can just query all the tags and then filter by hymn type, hymn number, and query param

skywalkerdude:
> Okay, so do you want to review this logic, merge into tag, and then I do the next piece with another PR etc and keep merging into tag... Or should I start committing each piece onto this PR and merge it all as one into tag?

Preferably we can review quickly enough that we'll only ever be 1 or two PRs ahead of master. If you approve my PR, I'll submit that into master, and then you can diffbase this off master instead of off my branch.

Hymns/Tags/TagStore.swift
finben: Wouldn't it? Lol because the Favorites tags are in a different entity right? The only problem with doing the PR this way is it's hard for me to check this stuff now lol...

finben: If they are all in the subclass FavoriteEntity then those objects would not be stored in the TagsEntity if I am right?

Hymns/Tags/TagStore.swift
finben: I would still need a function like his to take the hymnIdentifier and convert into string values for hymnType, hymn number, and query param right? I could take convertedKey out and filer on hymnIdentifier.hymnName but hymnType is not a string.

finben: Nevermind, I figured it out.

Hymns/Tags/TagStore.swift
skywalkerdude: Hmm, interesting. I think they still would be, since they're still using the same realm file.

But you can find out pretty easily via a unit test. Create an in-memory database, add stuff to it, and query to see what is and isn't in there: https://github.com/HymnalDreamTeam/Hymns-iOS/blob/master/HymnsTests/History/HistoryStoreRealmImplSpec.swift#L11

Hymns/Tags/TagStore.swift
finben: I guess the question here is also that we could probably just get rid of the second filter for it not being favorites. Since won't all favorites now only be stored in the FavoriteEntity and not the TagEntity. So there won't be any found when we query all object in TagEntity

Hymns/Tags/TagStore.swift
skywalkerdude: Okay, I guess you can try it. I have doubts that it'll work, because you're looking for TagEntities with an empty string as its tag.

As for hymn type not being a string, you can use hymnType.abbreviatedValue. But you have to see how hymnType is being stored to determine what you should call. It may work, or you may need to change the createPrimaryKey function to call hymnType.abbreviatedValue to make it work

Hymns/Tags/TagStore.swift
finben: Not really needed anymore. It was more relevant when there where a million functions in tagStore

Hymns/Tags/TagStore.swift
finben: It meant not including tags that where favorited. Also, not as relevant anymore since favorites obviously won't be included.

Hymns/Tags/TagStore.swift
finben: It looks like I was write. I think you can think of entities in Realm as tables. So they are two different tables. Take a look at my unit test and you will see that indeed favorites are not being stored there. Thanks for the tip to use unit tests like this.

Hymns/Tags/TagStore.swift
finben: It works as seen in my recent test. the reason is because the NSpredicate is using a CONTAINS for the filter. So even though they all have tags on the end of the string they still contain as part of their string this information.

Hymns/Tags/TagStore.swift
```diff
@@ -4,9 +4,8 @@ import RealmSwift
 import Resolver
 
 protocol TagStore {
-    func storeFavorite(_ entity: FavoriteEntity)
+    func storeRealmObject<T: Object>(_ entity: T)
```
skywalkerdude, line 7: What does this do? does making this function generic make both Favorites and Tags the same table?

Hymns/Tags/TagStore.swift
skywalkerdude: okay. Yeah, it looks like realm doesn't really support polymorphism: https://stackoverflow.com/questions/42809394/inheritance-with-swift-realm-confusion.

I think that's okay. Maybe we can just keep it as two tables then. In that case, we won't need a special "*_*_favorites*_*" tag, since they are different tables, and they'll never conflict

HymnsTests/Tags/TagsStoreRealmImplSpec.swift
```diff
@@ -0,0 +1,81 @@
...
+                }
+                describe("get the list of all favorites") {
+                    it("should be 0 when we query tagEntity") {
+
+                        let queryAllTags = target.querySelectedTags(tagSelected: nil)
+                        expect(queryAllTags).to(haveCount(0))
+                    }
+                    it("should be 3 when we query FavoriteEntity") {
+
+                        let queryAllFavorites = inMemoryRealm.objects(FavoriteEntity.self)
```
skywalkerdude, line 46: this should call target.something(...) instead of querying the realm db directly

skywalkerdude:
Okay. I think this mostly looks good, except I'm not sure why you did the "Make storing a realm object a generic function to work with either ta…" commit. Once we resolve that, I'll approve!

Hymns/Tags/TagStore.swift
skywalkerdude: actually, if we're just going to run with it being a separate table, should we just split it up between FavoritesStore and TagStore? Then things will be cleaner, since we are fully decoupling these into two separate features

finben:
The generic function only means that it can take either entity. But it will
still write to whichever entity is used in the function call. So no it
won't merge them. Yeah, we probably should just do two stores. That means
we probably won't use the generic function but instead just give them both
a specific function for store and delete.

Luke <notifications@github.com> schrieb am Do., 4. Juni 2020, 6:15 PM:

Hymns/Tags/TagStore.swift
finben: https://docs.swift.org/swift-book/LanguageGuide/Generics.html
Generics just allow you to not have to tie the function to one type in the parameters. So it could be called with any type as long as it conforms to the Object protocol here.

finben:
I changed base from tag to master as we discussed.

Hymns/Display/DisplayHymnViewModel.swift
```diff
@@ -18,36 +18,36 @@ class DisplayHymnViewModel: ObservableObject {
...
     private let analytics: AnalyticsLogger
     private let backgroundQueue: DispatchQueue
-    private let favoritesStore: FavoritesStore
     private let historyStore: HistoryStore
     private let identifier: HymnIdentifier
     private let mainQueue: DispatchQueue
     private let pdfLoader: PDFLoader
     private let repository: HymnsRepository
     private let storeInHistoryStore: Bool
+    private let favoriteStore: FavoriteStore
```
skywalkerdude, line 27: So I try to keep things in alphabetical order as much as I can, so can you put this back to where it was? (i.e. line 21)

Hymns/Display/DisplayHymnViewModel.swift
```diff
@@ -18,36 +18,36 @@ class DisplayHymnViewModel: ObservableObject {
...
          backgroundQueue: DispatchQueue = Resolver.resolve(name: "background"),
-         favoritesStore: FavoritesStore = Resolver.resolve(),
          hymnToDisplay identifier: HymnIdentifier,
          hymnsRepository repository: HymnsRepository = Resolver.resolve(),
          historyStore: HistoryStore = Resolver.resolve(),
          mainQueue: DispatchQueue = Resolver.resolve(name: "main"),
          pdfPreloader: PDFLoader = Resolver.resolve(),
-         storeInHistoryStore: Bool = false) {
+         storeInHistoryStore: Bool = false,
+         favoriteStore: FavoriteStore = Resolver.resolve()) {
```
skywalkerdude, line 40: ditto

Hymns/Display/DisplayHymnViewModel.swift
```diff
@@ -18,36 +18,36 @@ class DisplayHymnViewModel: ObservableObject {
...
         self.backgroundQueue = backgroundQueue
         self.currentTab = .lyrics(HymnLyricsView(viewModel: HymnLyricsViewModel(hymnToDisplay: identifier)).maxSize().eraseToAnyView())
-        self.favoritesStore = favoritesStore
         self.historyStore = historyStore
         self.identifier = identifier
         self.mainQueue = mainQueue
         self.pdfLoader = pdfPreloader
         self.repository = repository
         self.storeInHistoryStore = storeInHistoryStore
+        self.favoriteStore = favoriteStore
```
skywalkerdude, line 50: ditto

Hymns/Favorites/FavoriteStore.swift
```diff
@@ -67,16 +63,16 @@ class FavoritesStoreRealmImpl: FavoritesStore {
...
 extension Resolver {
-    public static func registerFavoritesStore() {
-        register(FavoritesStore.self) {
+    public static func registerFavoriteStore() {
+        register(FavoriteStore.self) {
             // https://stackoverflow.com/questions/28465706/how-to-find-my-realm-file
             var url = Realm.Configuration.defaultConfiguration.fileURL
             url?.deleteLastPathComponent()
-            url?.appendPathComponent("favorites.realm")
+            url?.appendPathComponent("tags.realm")
```
skywalkerdude, line 71: should be favorites.realm

Hymns/Infrastructure/Dependency Injection/AppDelegate+Injection.swift
```diff
@@ -28,7 +28,8 @@ extension Resolver: ResolverRegistering {
         registerConverters()
         registerHymnDataStore()
         registerHistoryStore()
-        registerFavoritesStore()
+        registerTagStore()
+        registerFavoriteStore()
```
skywalkerdude, line 32: FavoriteStore*

HymnsTests/Favorites/FavoriteStoreRealmImplSpec.swift
```diff
@@ -0,0 +1,52 @@
...
+                    target.storeFavorite(FavoriteEntity(hymnIdentifier: classic1151, songTitle: "Hymn 1151"))
+
+                    target
+                        .storeFavorite(FavoriteEntity(hymnIdentifier: newSong145, songTitle: "Hymn: Jesus shall reign where\\u2019er the sun"))
+
+                    target
+                        .storeFavorite(FavoriteEntity(hymnIdentifier: cebuano123, songTitle: "Naghigda sa lubong\\u2014"))
+                }
+                describe("get the list of all favorites") {
+                    it("should return a bool when we query if it is favorited") {
```
skywalkerdude, line 39: "cebuano 123 should be favorited"

HymnsTests/Tags/TagsStoreRealmImplSpec.swift
```diff
@@ -0,0 +1,57 @@
...
+                    target
+                        .storeTag(TagEntity(hymnIdentifier: cebuano123, songTitle: "Naghigda sa lubong\\u2014", tag: "Table"))
+                }
+                describe("getting all tags") {
+                    it("should contain the stored songs sorted by last-stored") {
+                        let resultsOfQuery = target.querySelectedTags(tagSelected: nil)
+                        expect(resultsOfQuery).to(haveCount(3))
+                    }
+                }
+                describe("getting one hymns tags after storing multiple tags for that hymn") {
```
skywalkerdude, line 43: hymn's*

HymnsTests/Tags/TagsStoreRealmImplSpec.swift
```diff
@@ -0,0 +1,57 @@
...
+                    target.storeTag(TagEntity(hymnIdentifier: classic1151, songTitle: "Hymn 1151", tag: "Christ"))
+
+                    target
+                        .storeTag(TagEntity(hymnIdentifier: newSong145, songTitle: "Hymn: Jesus shall reign where\\u2019er the sun", tag: "Bread and wine"))
+
+                    target
+                        .storeTag(TagEntity(hymnIdentifier: cebuano123, songTitle: "Naghigda sa lubong\\u2014", tag: "Table"))
+                }
+                describe("getting all tags") {
+                    it("should contain the stored songs sorted by last-stored") {
```
skywalkerdude, line 38: the description states that it should be sorted but you don't have any assertions on the order of the tags

HymnsTests/Tags/TagsStoreRealmImplSpec.swift
```diff
@@ -0,0 +1,57 @@
...
+            }
+            afterEach {
+                // Don't worry about force_try in tests.
+                // swiftlint:disable:next force_try
+                try! inMemoryRealm.write {
+                    inMemoryRealm.deleteAll()
+                }
+                inMemoryRealm.invalidate()
+            }
+            context("store a few tags") {
```
skywalkerdude, line 27: Can you also test the deleting tags as well?

Hymns/Favorites/FavoriteEntity.swift
```diff
@@ -0,0 +1,19 @@
...
+        super.init()
+    }
+
+    init(hymnIdentifier: HymnIdentifier, songTitle: String) {
+        super.init(hymnIdentifier: hymnIdentifier, songTitle: songTitle, tag: Self.favoriteTag)
+    }
+
+    static func createPrimaryKey(hymnIdentifier: HymnIdentifier) -> String {
+        return TagEntity.createPrimaryKey(hymnIdentifier: hymnIdentifier, tag: favoriteTag)
+    }
```
skywalkerdude, line 18: Maybe we can revert this file too, instead of piggyback off of TagEntity. It seems like this current behavior is just lucky and isn't technically supported by Realm, so i'm scared that they'll change this undefined behavior somewhere down the line and our stuff will break.

Hymns/Infrastructure/Dependency Injection/AppDelegate+Injection.swift
finben: ? Hmmm isn't that right

Hymns/Favorites/FavoriteEntity.swift
finben: Hahaha lol! We have gone full circle.... xD   Now there is basically no need of even having a favoriteTag. We might as well go back to how we had it where just delete objects when they are unfavorited and not even store the variable favorite tag? Is that what you had in mind too?

finben: Addressed in recent commit

HymnsTests/Tags/TagsStoreRealmImplSpec.swift
finben: Addressed in recent commit

Hymns/Infrastructure/Dependency Injection/AppDelegate+Injection.swift
skywalkerdude: well the class is "FavoriteStore", so the function should be "registerFavoriteStore" and not "registerFavoritesStore"

skywalkerdude: Are you able to fix this and then submit? Or do you need another approval after that?

finben:
Huh. Isn't that a delete comment next to the line removing that?

Luke <notifications@github.com> schrieb am Do., 4. Juni 2020, 10:54 PM:

skywalkerdude:
ohh you did fix it. oops. okay. lgtm!

skywalkerdude:
@finben it looks like you rebased this branch instead of squashing it: 
Screen Shot 2020-06-05 at 9 47 20 AM

I guess it's not a huge deal, but I think to keep the git history clean, we should try to squash these commits instead of rebasing them

PR #241, opened 2020-06-05
Tag view model

finben:
I added tests to show both are working as expected.


Hymns/Tags/TagStore.swift
```diff
@@ -65,6 +66,19 @@ class TagStoreRealmImpl: TagStore {
     func queryTagsForHymn(hymnIdentifier: HymnIdentifier) -> Results<TagEntity> {
         return realm.objects(TagEntity.self).filter(NSPredicate(format: "primaryKey CONTAINS[c] %@", ("\(hymnIdentifier.hymnType):\(hymnIdentifier.hymnNumber):\(hymnIdentifier.queryParams ?? [String: String]())")))
     }
+
+    //A workaround to the fact that Realm does not yet support querying unique values. Force cast should never fail since the realm object will always be a string and castable into a string array.
+    //swiftlint:disable force_cast
```
skywalkerdude, line 71: It seems like this is supported: https://stackoverflow.com/a/49523428/1907538

Hymns/Tags/TagStore.swift
```diff
@@ -9,6 +9,7 @@ protocol TagStore {
     func observeTagStatus(hymnIdentifier: HymnIdentifier, action: @escaping (Bool) -> Void) -> Notification
     func querySelectedTags(tagSelected: String?) -> Results<TagEntity>
     func queryTagsForHymn(hymnIdentifier: HymnIdentifier) -> Results<TagEntity>
+    func queryUniqueTags() -> [TagEntity]
```
skywalkerdude, line 12: If you're just getting the tags, I'd say this should probably be a `[String]` not a `[TagEntity]`, since you don't really care about any of the other stuff

skywalkerdude, line 12: this is just my personal preference, but I prefer "get*" over "query*". So "getAllTags" or "getTagsForHymn". but up to you

Hymns/Tags/TagListViewModel.swift
```diff
@@ -0,0 +1,56 @@
+import Combine
+import SwiftUI
+import RealmSwift
+import Resolver
+
+class TagListViewModel: ObservableObject {
```
skywalkerdude, line 6: This should probably be in the "Browse" group, not in "Tags"

Hymns/Tags/TagListView.swift
```diff
@@ -0,0 +1,29 @@
+import SwiftUI
+import Resolver
+
+struct TagListView: View {
```
skywalkerdude, line 4: This should probably be in the "Browse" group, not in "Tags"

Hymns/Tags/TagSubList.swift
```diff
@@ -0,0 +1,24 @@
...
+            List(self.viewModel.tags) { tag in
+                NavigationLink(destination: tag.destinationView) {
+                    SongResultView(viewModel: tag)
+                }
+            }
+        }.onAppear {
+            self.viewModel.fetchTagsByTags(self.tagSelected.title)
+        }
+    }
+}
```
skywalkerdude, line 24: So I have this thing already called BrowseResultsListView and BrowseResultsListViewModel. I would like to reuse that as much as possible. I know right now it has code to fetch Category/Subcategory, but if we could make that more generic and use that, that'd be  preferable to creating this special sublist that basically just does the same thing

skywalkerdude: oh I have an idea. What if we just overrode the init functions.

So you'd create your own init function:
```
    init(paramsThatYouNeed, dataStore: HymnDataStore = Resolver.resolve(),
         mainQueue: DispatchQueue = Resolver.resolve(name: "main")) {

       // fill in the title and results

    }
```

finben: I'll try it. I've been looking at it the past hour or so. I was going a route of trying to use protocols and generics. But let me see if this solution is doable. Seems like it would be nice.

finben: I couldn't get that to work because the @ObservedObject still needs a type specified before the init function

Hymns/Tags/TagListView.swift
finben: Moded

Hymns/Tags/TagStore.swift
finben: After changing the function it now returns a Results<TagEntity> like all the other get/query functions. The problem with making it a string is then it won't fit into the same variable @Published var tags  at the viewModel level. Should I still make it a [String]

Hymns/Tags/TagStore.swift
finben: Nice! Great find. I made the change

Hymns/Tags/TagSubList.swift
finben: Can we just note this for a refactor later lol :/

Hymns/Tags/TagStore.swift
skywalkerdude: I don't understand what you mean by "won't fit into the same variable". Can't you just change the view model be Result<String> instead of Result<TagEntity>?

Hymns/Tags/TagSubList.swift
skywalkerdude: Are you referring to the protocol approach? I don't think that'll work. Did you try overriding the init function?

finben:
I am referring to the init, but I tried both. I think the protocol with
generics could work.

If tags and categories where the same type it would be much easier

Luke <notifications@github.com> schrieb am Fr., 5. Juni 2020, 2:52 PM:

Hymns/Tags/TagStore.swift
finben: Do you still want me to try to do something here or are we good to merge this PR?

Hymns/Tags/TagSubList.swift
finben: Nice work!

skywalkerdude:
Are you able to make `func queryUniqueTags() -> [TagEntity]` return `func queryUniqueTags() -> [String]` instead?

Hymns/Browse/BrowseCategoriesView.swift
```diff
@@ -18,7 +18,7 @@ struct BrowseCategoriesView: View {
 
             return List(categories) { category in
                 CategoryView(viewModel: category)
-            }.eraseToAnyView()
+            }.listStyle(GroupedListStyle()).eraseToAnyView()
```
skywalkerdude, line 21: What does GroupedListStyle do?

Hymns/Browse/BrowseResultsListViewModel.swift
```diff
@@ -9,6 +9,14 @@ class BrowseResultsListViewModel: ObservableObject {
 
     private var disposables = Set<AnyCancellable>()
 
+    init(tag: String, tagStore: TagStore = Resolver.resolve()) {
+        self.title = tag
+        songResults = tagStore.getSelectedTags(tagSelected: tag).map { tagEntity -> SongResultViewModel in
+            let hymnIdentifier = HymnIdentifier(tagEntity.hymnIdentifierEntity)
+            return SongResultViewModel(title: tagEntity.songTitle, destinationView: DisplayHymnView(viewModel: DisplayHymnViewModel(hymnToDisplay: hymnIdentifier)).eraseToAnyView())
+        }
+    }
```
skywalkerdude, line 18: I know I wrote this code, but could you help me add a unit test for this?

Hymns/Browse/TagListView.swift
```diff
@@ -0,0 +1,29 @@
...
+    init(viewModel: TagListViewModel = Resolver.resolve()) {
+        self.viewModel = viewModel
+    }
+
+    var body: some View {
+        VStack {
+            List(viewModel.tags, id: \.self) { tag in
+                NavigationLink(destination: BrowseResultsListView(viewModel: BrowseResultsListViewModel(tag: tag))) {
+                    Text(tag)
+                }
```
skywalkerdude, line 17: I forgot to add a TODO here for the empty state, for when viewModel.tags is empty

Hymns/Tags/TagStore.swift
skywalkerdude: oh, yeah. Could you try to get it to return [String] instead of [TagEntity]?

HymnsTests/Tags/TagListViewModelSpec.swift
```diff
@@ -0,0 +1,55 @@
...
+import Mockingbird
+import Nimble
+import RealmSwift
+@testable import Hymns
+
+class TagListViewModelSpec: QuickSpec {
+    override func spec() {
+        describe("using an in-memory realm") {
+            var inMemoryRealm: Realm!
+            var tagStore: TagStoreRealmImpl!
```
skywalkerdude, line 12: Can you mock this instead?

`var tagStore: TagStoreMock!`

skywalkerdude:
ok i fixed it. there's still 1 open question

finben:
Is this the open question because it is already addressed? I don't see any other questions.
"Are you able to make func queryUniqueTags() -> [TagEntity] return func queryUniqueTags() -> [String] instead?"

Hymns/Favorites/FavoritesView.swift
```diff
@@ -23,7 +23,7 @@ struct FavoritesView: View {
                     NavigationLink(destination: favorite.destinationView) {
                         SongResultView(viewModel: favorite)
                     }
-                }.resignKeyboardOnDragGesture()
+                }.listStyle(GroupedListStyle()).resignKeyboardOnDragGesture()
```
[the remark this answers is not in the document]
skywalkerdude: this looks like it's still there

skywalkerdude:
`GroupedListSyle` is still there in https://github.com/HymnalDreamTeam/Hymns-iOS/pull/241/files#diff-24ea0213b6cad5d70baae65823e29643

Issue #231, opened 2020-06-02
Navigating back to the home screen always goes to the home tab


skywalkerdude:
@finben it looks like you added this PR: https://github.com/HymnalDreamTeam/Hymns-iOS/pull/118

Do you remember what the issue that this fix is for? I've found that it breaks navigation because of
```
            }.onAppear {
                self.selectedTab = .home
            }
```

So when you navigate back from any other screen, the `onAppear` will force it to go back to the home tab instead of the tab you were previously on

skywalkerdude:
See video:

[video-1590038334.mp4.zip](https://github.com/HymnalDreamTeam/Hymns-iOS/files/4719492/video-1590038334.mp4.zip)

finben:
Yeah, the problem is this is a fix for that first icon being lower than the others. 
![Simulator Screen Shot - iPhone 11 Pro Max - 2020-06-05 at 10 48 26](https://user-images.githubusercontent.com/15247726/83896697-2d58bd00-a71a-11ea-911d-47818a99dad6.png)

PR #243, opened 2020-06-05
Fix navigation issue while keeping icons aligned

finben:
Addresses #231 by basically only making the state change during the first time this view ever appears.


finben:
I tried a bunch of other options too like onReceive with a publisher etc. None, of them worked :/ this works great though.

skywalkerdude:
awesoome!

PR #244, opened 2020-06-05
Create Scripture views, but haven't hooked everything up yet

skywalkerdude:


Hymns/Infrastructure/Data/Models/Book.swift
```diff
@@ -0,0 +1,68 @@
+enum Book: String {
```
skywalkerdude, line 1: @finben everything should be right, but please double-check my spelling!

Issue #175, opened 2020-05-22
Generic empty state for when web view fails to load


skywalkerdude:
Web views may fail to load. Currently we only use it to show the privacy policy, but there could be other stuff in the future.

![Simulator Screen Shot - iPhone 8 - 2020-05-22 at 00 33 38](https://user-images.githubusercontent.com/1427524/82643204-fafc7b00-9bc3-11ea-870f-e74a9cd84143.png)

Emyeh:
Would we be able to detect if it was a connection issue? @skywalkerdude

skywalkerdude:
I don't think so, at least not right now that I know of

Emyeh:
Added a generic error state for both iOS and Android: https://www.figma.com/file/7H5ApjzuVuJL3OP7FvE2qe/h-app?node-id=1418%3A6216

PR #253, opened 2020-06-06
create error states and add to web view

skywalkerdude:


skywalkerdude:
closes https://github.com/HymnalDreamTeam/Hymns-iOS/issues/175

PR #255, opened 2020-06-07
add error state for browse results

skywalkerdude:


skywalkerdude:
closes https://github.com/HymnalDreamTeam/Hymns-iOS/issues/175

PR #258, opened 2020-06-08
Backend to store tag colors. Tested in a test as well

finben:


finben:
Ignore the branch name... This is basically just to add storing color values using an enum to TagEntity

Hymns/Tags/TagEntity.swift
```diff
@@ -27,3 +33,7 @@ class TagEntity: Object, Identifiable {
         return ("\(hymnIdentifier.hymnType):\(hymnIdentifier.hymnNumber):\(hymnIdentifier.queryParams ?? [String: String]()):\(tag)")
     }
 }
+
+enum TagColor: Int {
+        case blue, green, yellow, red
```
skywalkerdude, line 38: nit: indenting

Hymns/Tags/TagEntity.swift
```diff
@@ -6,17 +6,23 @@ class TagEntity: Object, Identifiable {
     @objc dynamic var hymnIdentifierEntity: HymnIdentifierEntity!
     @objc dynamic var songTitle: String!
     @objc dynamic var tag: String!
+    @objc private dynamic var privateTagColor: Int = TagColor.blue.rawValue
```
skywalkerdude, line 9: Can you add `// https://stackoverflow.com/questions/29123245/using-enum-as-property-of-realm-model` as a comment on this field

Hymns/Tags/TagEntity.swift
skywalkerdude, line 38: I think you should have a default case also

Hymns/Tags/TagEntity.swift
skywalkerdude, line 9: make default case .default instead of blue

Hymns/Tags/TagStore.swift
```diff
@@ -7,7 +7,7 @@ protocol TagStore {
     func storeTag(_ entity: TagEntity)
     func deleteTag(primaryKey: String, tag: String)
     func getSongsByTag(_ tag: String) -> [SongResultViewModel]
-    func getTagsForHymn(hymnIdentifier: HymnIdentifier) -> [String]
+    func getTagsForHymn(hymnIdentifier: HymnIdentifier) -> ([String], [TagColor])
```
skywalkerdude, line 10: Instead of a Tuple of Arrays, can you just have an array of tuples?

Hymns/Tags/TagEntity.swift
finben: I could not just add default since it is an enum keyword that doesn't work here but only in switch statements. I added a none as seen here https://useyourloaf.com/blog/enum-raw-values-and-failable-initializers/
Does that still accomplish what you are after?

finben: fixed

Hymns/Tags/TagEntity.swift
finben: done

Hymns/Tags/TagStore.swift
finben: Addressed in recent commit

finben:
I am not sure what you mean by store the migration code as a class? I just set one up following the tutorial from the Realm site you linked which is in appDelegate. One thing to consider, is don't we just want migration code for our entire realm database including both TagEntity and FavoritesEntity in one place since they are simply two tables of the same database? There should only be one realm file on the users phone for both of them so it makes since they would share a migration file right?

Hymns/Tags/TagEntity.swift
skywalkerdude: yup, that works!

Hymns/Tags/TagStore.swift
```diff
@@ -55,14 +55,17 @@ class TagStoreRealmImpl: TagStore {
...
+        let filteredObject = realm.objects(TagEntity.self)
+            .filter(NSPredicate(format: "primaryKey CONTAINS[c] %@", ("\(hymnIdentifier.hymnType):\(hymnIdentifier.hymnNumber):\(hymnIdentifier.queryParams ?? [String: String]())")))
+        let tags: [(tagName: String, tagColor: TagColor)] = filteredObject.map { entity -> (String, TagColor) in
+            (tagName: entity.tag, tagColor: entity.tagColor)
         }
         return tags
+//        let colors: [TagColor] = filteredObject.map { entity -> TagColor in
+//            entity.tagColor
+//        }
+//        return (tags, colors)
```
skywalkerdude, line 68: ?

Hymns/AppDelegate.swift
```diff
@@ -22,6 +23,28 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+            // Set the block which will be called automatically when opening a Realm with
+            // a schema version lower than the one set above
+            migrationBlock: { migration, oldSchemaVersion in
+                // We haven’t migrated anything yet, so oldSchemaVersion == 0
+                if oldSchemaVersion < 1 {
+                    migration.enumerateObjects(ofType: TagEntity.className()) { _, newObject in
+                        newObject!["privateTagColor"] = 0
+                    }
+                    // Nothing to do!
+                    // Realm will automatically detect new properties and removed properties
```
skywalkerdude, line 42: I'm a bit confused with the statement `Realm will automatically detect new properties and removed properties`. What does that mean? Does that mean we don't need to initialize the new properties?

Hymns/AppDelegate.swift
```diff
@@ -22,6 +23,28 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+            // version (if you've never set a schema version before, the version is 0).
+            schemaVersion: 1,
+
+            // Set the block which will be called automatically when opening a Realm with
+            // a schema version lower than the one set above
+            migrationBlock: { migration, oldSchemaVersion in
+                // We haven’t migrated anything yet, so oldSchemaVersion == 0
+                if oldSchemaVersion < 1 {
+                    migration.enumerateObjects(ofType: TagEntity.className()) { _, newObject in
+                        newObject!["privateTagColor"] = 0
```
skywalkerdude, line 39: Can you use the enum value instead of a hardcoded 0? `TagColor.none.rawValue`

Hymns/AppDelegate.swift
```diff
@@ -22,6 +23,28 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+                    }
+                    // Nothing to do!
+                    // Realm will automatically detect new properties and removed properties
+                    // And will update the schema on disk automatically
+                }
+        })
+        // Tell Realm to use this new configuration object for the default Realm
+        Realm.Configuration.defaultConfiguration = config
```
skywalkerdude, line 47: Hmmm, we have multiple Realm files in our app. Do you know how this would work? This looks like it's making this config the default config across the board for all the realm files in our app, so wouldn't this also affect the history file as well?
https://stackoverflow.com/a/54249204

Hymns/Tags/TagEntity.swift
skywalkerdude: hmmm, actually, what if we made TagColor an optional value? I'm not sure if that would be better or worse or would make no difference. What do you think would be the pros and cons of that versus creating a special "none" enum value?

finben: I think it would make no difference. If you want I can do it though. I don't think it would make a difference since we will always have a tagColor stored. I think the only pro is it may look slightly better in code. But the con is we have to deal with an optional, unwrap it etc.. 

Let me know what you think

Hymns/AppDelegate.swift
finben: I mean, I think it would be the same as now actually in a sense. Because right now they are still all being rooted from the same Realm.Configuration.defaultConfiguration anyways. We are just calling the .fileUrl method on the existing default configuration. 

But yeah, to answer your question it would affect all of our realm files. Which may be a good thing because there will just be one place for all migration versions. So it's not like we have them staged at various versions. Idk.

Hymns/AppDelegate.swift
finben: Sorry, lol thats not an accurate comment. I went ahead and deleted it. We definitely do need to initialize new properties.

Hymns/Tags/TagEntity.swift
skywalkerdude: ok. It's probably simpler then to just keep the "none" enum value, since we can just use a switch statement instead of dealing with wrapping/unwrapping optionals!

Hymns/AppDelegate.swift
skywalkerdude: > But yeah, to answer your question it would affect all of our realm files. Which may be a good thing because there will just be one place for all migration versions.
I think having all the migration code in one place is good, but having a schema upgrade for one realm file affect all the realm files, I don't think is a good thing.
 
My concern is like, say we change Tags to version 2, now do we need to provide a schema version for the history store as well? It's confusing because we are using one schema version to keep track of potentially multiple realm files.

So, let's say we increment the schema version from 1 to 2. Now both the History and Tags will run the migration code. Best case here, it just waste CPU cycles, worse case, there are funky race conditions because they are both trying to increment the TagEntity. I suspect the former will happen because they are different realm files.

I'd be more comfortable if you created a configuration per realm file, and just had the migration code specific to that realm file. In this case, you can just create the configuration file for Tags, since we haven't had the need to do any migrations for History or Favorites.

To your point about having all migration code in the same place, it's a very good point. We can create a `Migrations` class in `Infrastructure/Data` and put all the migration code there. That way we can unit test also!

```
let config = Realm.Configuration(
    // Set the new schema version. This must be greater than the previously used
    // version (if you've never set a schema version before, the version is 0).
    schemaVersion: 1,

    // Set the block which will be called automatically when opening a Realm with
    // a schema version lower than the one set above
    migrationBlock: { migration, oldSchemaVersion in
        // We haven’t migrated anything yet, so oldSchemaVersion == 0
        if oldSchemaVersion < 1 {
            Migrations.performTagsMigration()
        }
})

let realm = try! Realm(fileURL: url!, configuration: config)
return TagStoreRealmImpl(realm: realm) as TagStore
```

https://exceptionshub.com/realm-migrations-in-swift.html

finben: I made some changes. Please check my work though. Not sure if I am doing this right.

skywalkerdude: lol bro. I realized that we haven't launched Tags to anybody yet... So we literally don't even need a migration plan. Migrations are for people who have an older version and upgrade, but nobody has the older version >.>

I just added some stub migration configs for each of the tag stores, but we don't need any real migration code at all lol. All we need to do is uninstall the app from our emulator when the version has a mismatch

PR #260, opened 2020-06-08
Added send feedback. You can't test from simulator fyi.

finben:
The button will be greyed out on simulator since you can't send email from there. If you want to test has to be on a device. I sent a feedback email from Johanna's iphone. can you make sure you got it?


finben:
Also, because I had to test on a real device I had to update development team to my own for it to test. But that changed the team name on xcodeproj but I still had to commit that file for the new files added. Can you change the development team back to yours on this branch and then commit over my commit? I can't change it back to yours from my xcode.

skywalkerdude:
I don't have an automated way of changing it either. I had to manually vim into the file and change it

Hymns/Settings/MailFeedbackView.swift
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    func makeCoordinator() -> Coordinator {
+        return Coordinator(presentation: presentation,
+                           result: $result)
+    }
+
+    func makeUIViewController(context: UIViewControllerRepresentableContext<MailFeedbackView>) -> MFMailComposeViewController {
+        let mailVC = MFMailComposeViewController()
+        let randomIssueNum = Int.random(in: 1000000000...9999999999)
```
skywalkerdude, line 43: lol

Hymns/Settings/FeedbackView.swift
```diff
@@ -0,0 +1,26 @@
+import SwiftUI
+import MessageUI
+
+struct FeedbackView: View {
+
+    @State var result: Result<MFMailComposeResult, Error>?
```
skywalkerdude, line 6: What do you do with this result?

Would it be possible to show like a simple toast message telling them that the email succeeded or failed?

We can do this as a follow-up or create an issue for it, since it look like creating a toast message is non-trivial

skywalkerdude: found some pointers here: https://stackoverflow.com/questions/56550135/swiftui-global-overlay-that-can-be-triggered-from-any-view

finben: Let's do it as a follow up lol.. I'll log an issue. And the result is used between the UIViewRepresentable and Coordinator as a binding variable passed to both of them.

Issue #254, opened 2020-06-06
Implement PDF error state


skywalkerdude:
Related to https://github.com/HymnalDreamTeam/Hymns-iOS/issues/228.

We need to show the generic error state (added in https://github.com/HymnalDreamTeam/Hymns-iOS/pull/253) when a PDF fails to load for an unknown reason.

Note: If possible, please do not merely copy what I did for the web view. That only checks if the url is invalid or not. This is okay for webviews because if the url is valid but doesn't point to anything, the WebView will show a 404 page, which is an error screen in and of itself. Pdf viewer has no such functionality. If the url is valid but the pdf viewer fails to render, then it'll still just show a blank page. Therefore, we need to DETECT when a pdf fails to load and show the error view, so it is a little bit more involved than merely checking the url's validity.

Emyeh:
If the music sheet is not available can we remove the tab?

skywalkerdude:
@Emyeh it's not that the sheet music is not available. It's that there is a url, but the url doesn't load for some reason. We won't be able to hide it in that case because we won't know that the url is wrong at the time of rendering the tabs.

Emyeh:
[Error.pdf](https://github.com/HymnalDreamTeam/Hymns-iOS/files/4762374/Error.pdf)

Attached a pdf export

finben:
Done! Thanks @Emyeh

PR #263, opened 2020-06-09
Add a PDF error state as an example

finben:
So I think the best way to do this is actually have an error state that is a PDF. That way it can catch when there is no pdf document and then be displayed in its place. Obviously right now the error state is just an outline for a conference lol as an example. I just want to show you and if you agree maybe we can get an error state for this from Emily?


Hymns/Common/PDFViewer.swift
```diff
@@ -22,6 +22,13 @@ struct PDFViewer: UIViewRepresentable {
         } else {
             pdfView.document = PDFDocument(url: url)
         }
+
+        //Display error state
+        if pdfView.document == nil {
+            if let fileURL = Bundle.main.url(forResource: "pdfErrorState", withExtension: "pdf") {
+                pdfView.document = PDFDocument(url: fileURL)
+            }
+        }
```
skywalkerdude, line 31: Display error state when url is niil also (TODO on line 16)

skywalkerdude:
Fixes https://github.com/HymnalDreamTeam/Hymns-iOS/issues/254

finben:
Migrated issue to #273

Issue #238, opened 2020-06-04
Designs for tags


finben:
@Emyeh When you get the chance could you provide some designs for tagging hymns. Right now I have this coded up but I am sure this is not what we want lol 
![Simulator Screen Shot - iPhone 11 - 2020-06-04 at 12 33 06](https://user-images.githubusercontent.com/15247726/83791790-d393bc80-a65f-11ea-9fb2-3e906da35bbe.png)

Emyeh:
You guys are too fast hahah. I'll work on the design for this.

Emyeh:
Updated design for tag ios: 
https://www.figma.com/file/7H5ApjzuVuJL3OP7FvE2qe/h-app?node-id=647%3A3155

Let me know if you have any questions.

Emyeh:
Updated the iOS design for multiple tags. Let me know what you think. Here is the prototype for how the interaction works adding and removing a tag: https://www.figma.com/proto/7H5ApjzuVuJL3OP7FvE2qe/h-app?node-id=1468%3A1028&scaling=min-zoom

finben:
Look great!! Conversation is mostly resolved and moved to #257

PR #266, opened 2020-06-11
Tag sheet with color selection for tags

finben:
Obviously, this is another giant PR. I can split it up into smaller ones that we can merge in but I just want to save all my work onto a PR.


finben:
Can you take a look at TagStore. I am wanting to change the getTagsForHymn function back to returning  Results<TagEntity> instead of the array tuple. The reason I am wanting to do this is twofold.

1. When returning the array storing the tuple when the realm updates it wasn't updating the tags on the tag sheet view when I would click "Add". I think that is because it couldn't take advantage of the Realm notifications when the object would change. 

2. I couldn't ForEach over the array of tuples. SwiftUI does not find that to be identifiable. So I would still have to map them out in the viewModel to objects. 

Anyways, like I mention above I can split this up into smaller PR's again but can you give me feedback on this question first?

skywalkerdude:
So I found this pretty cool thing where you can call `publisher` and it'll make a self-updating Combine publisher:
https://realm.io/docs/swift/latest/api/Protocols/RealmCollection.html#/s:10RealmSwift0A10CollectionPA2A0A12SubscribableRzrlE9publisherAA10PublishersO5ValueVy_xGvp
https://github.com/realm/realm-cocoa/blob/v5.0.2/RealmSwift/Combine.swift#L260-L262

I'm using this right now to refactor favorites, but i think you can use this to accomplish what you want

skywalkerdude:
> 2\. I couldn't ForEach over the array of tuples. SwiftUI does not find that to be identifiable. So I would still have to map them out in the viewModel to objects.

If you can't ForEach over an array of tuples, feel free to create another struct that contains the tag string and the tag color and make that identifiable

skywalkerdude:
To make this smaller, you can start by not including the colors and just make everything the `.none` color for now. I don't think you created any way of displaying the colors in `TagListView` anyway.

skywalkerdude:
closes https://github.com/HymnalDreamTeam/Hymns-iOS/issues/238

finben:
Okay, the way I want to do this is like what we did before. We can do the PR for the backend models and tests first and then merge in the front end stuff. Right now this PR contains both. But I just created another PR #272 for the backend and VM changes that are going to be needed. After we merge that in then I will make a PR for just the front end. But don't delete this branch in the mean time.

skywalkerdude:
@finben close this?

PR #272, opened 2020-06-11
Tag sheet VM

finben:
All of the backend stuff for tag sheets with color and everything. We can get this one fixed up and merged first and then put my other front end PR in after that.


finben:
Tag colors are included for the tag sheet now but I will add the support for the tag color on the view model list in another PR later. The reason is I still have questions about the implementation that I am waiting on a response from Emily for.

Hymns/Tags/TagEntity.swift
```diff
@@ -34,8 +34,48 @@ class TagEntity: Object, Identifiable {
...
 
 enum TagColor: Int {
     case none, blue, green, yellow, red
 }
+
+extension TagColor {
+    var background: UIColor {
+        switch self {
+        case .none:
+            return UIColor.white
```
skywalkerdude, line 55: Probably use systembackground

Hymns/Tags/TagEntity.swift
```diff
@@ -34,8 +34,48 @@ class TagEntity: Object, Identifiable {
...
+            return UIColor(red: 255/255, green: 209/255, blue: 0/255, alpha: 0.2)
+        case .red:
+            return UIColor(red: 255/255, green: 0, blue: 31/255, alpha: 0.2)
+        }
+    }
+
+    var foreground: UIColor {
+        switch self {
+        case .none:
+            return UIColor.white
```
skywalkerdude, line 70: use primary color

Hymns/Tags/TagStore.swift
```diff
@@ -72,6 +75,7 @@ class TagStoreRealmImpl: TagStore {
     }
 }
 
+
```
skywalkerdude, line 78: fix swiftlint

HymnsTests/Tags/TagsStoreRealmImplSpec.swift
```diff
@@ -37,14 +37,28 @@ class TagStoreRealmImplSpec: QuickSpec {
...
+                                case .finished:
+                                    finished.fulfill()
+                                }
+                                return
+                            }, receiveValue: { entities in
+                                value.fulfill()
+                                expect(entities).to(contain([TagEntity(hymnIdentifier: classic1151, songTitle: "Hymn 1151", tag: "Christ", tagColor: .blue), TagEntity(hymnIdentifier: classic1151, songTitle: "Hymn 1151", tag: "Is", tagColor: .red), TagEntity(hymnIdentifier: classic1151, songTitle: "Hymn 1151", tag: "Life", tagColor: .red), TagEntity(hymnIdentifier: classic1151, songTitle: "Hymn 1151", tag: "Peace", tagColor: .blue)]))
+                            })
+                        self.wait(for: [failure, finished, value], timeout: testTimeout)
+                        cancellable.cancel()
```
skywalkerdude, line 61: This looks great!

Hymns/Tags/TagSheetViewModel.swift
```diff
@@ -0,0 +1,95 @@
...
+
+    init(title: String, color: TagColor) {
+        self.title = title
+        self.color = color
+    }
+}
+
+extension TagMeta: Hashable {
+    static func == (lhs: TagMeta, rhs: TagMeta) -> Bool {
+        lhs.title == rhs.title
```
skywalkerdude, line 83: Should we not also compare the tag color?

Hymns/Tags/TagSheetViewModel.swift
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    func hash(into hasher: inout Hasher) {
+        hasher.combine(title)
```
skywalkerdude, line 87: ditto

Hymns/Tags/TagSheetViewModel.swift
```diff
@@ -0,0 +1,95 @@
...
+    }
+}
+
+extension Resolver {
+    public static func registerTagSheetViewModel() {
+        register {TagSheetViewModel(hymnToDisplay: Resolver.resolve())}.scope(graph)
```
skywalkerdude, line 93: Does this work? How are you resolving the hymn identifier?

Hymns/Tags/TagSheetViewModel.swift
```diff
@@ -0,0 +1,95 @@
...
+    func addTag(tagName: String, tagColor: TagColor) {
+        self.tagStore.storeTag(TagEntity(hymnIdentifier: self.identifier, songTitle: self.title, tag: tagName, tagColor: tagColor))
+    }
+
+    func deleteTag(tagTitle: String, tagColor: TagColor) {
+        self.tagStore.deleteTag(primaryKey: TagEntity.createPrimaryKey(hymnIdentifier: self.identifier, tag: tagTitle), tag: tagTitle)
+    }
+}
+
+class TagMeta: Identifiable {
```
skywalkerdude, line 70: To be consistent with our other UI models, I would call this UiTag instead of TatgMeta

Hymns/Tags/TagSheetViewModel.swift
```diff
@@ -0,0 +1,95 @@
...
+            .sink(receiveValue: { results in
+                self.tags = results
+            }).store(in: &disposables)
+    }
+
+    func addTag(tagName: String, tagColor: TagColor) {
+        self.tagStore.storeTag(TagEntity(hymnIdentifier: self.identifier, songTitle: self.title, tag: tagName, tagColor: tagColor))
+    }
+
+    func deleteTag(tagTitle: String, tagColor: TagColor) {
```
skywalkerdude, line 65: Here, one thing is called tagName and one thing is called tagTitle. Are these different things?

Hymns/Tags/TagSheetViewModel.swift
```diff
@@ -0,0 +1,95 @@
...
+                    guard let self = self, let hymn = hymn, !hymn.lyrics.isEmpty else {
+                        return
+                    }
+                    let title: Title
+                    if self.identifier.hymnType == .classic {
+                        title = "Hymn \(self.identifier.hymnNumber)"
+                    } else {
+                        title = hymn.title.replacingOccurrences(of: "Hymn: ", with: "")
+                    }
+                    self.title = title
```
skywalkerdude, line 43: We're going to need to change this at some point, so could you extract this into a common function or something?

Maybe you can make the `title` field in UiHymn a computed property and stick this in there. That way both DisplayHymnViewModel and this view model can both just call it

Hymns/Tags/TagSheetViewModel.swift
finben: I'm not sure. It works fine for me as it is. What exactly is the advantage of doing so? Isn't this just necessary so that we can use ForEach on the array of this object? If so I don't think it is necessary since we never would ForEach just the label colors. But if there are performance benefits then we could do it.

Hymns/Tags/TagSheetViewModel.swift
finben: Lol good question :P I'm not really sure ha. It seems to work fine now. Maybe when we try to use it in a test it won't work? Idk

Hymns/Tags/TagSheetViewModel.swift
finben: Nice! done in recent commit

Hymns/Tags/TagSheetViewModel.swift
finben: Same, I change their names to be the same, and use tagTitle

Hymns/Tags/TagSheetViewModel.swift
finben: addressed

Hymns/Tags/TagEntity.swift
finben: I had to use label for the foreground since primary is not a UI Color

Hymns/Tags/TagEntity.swift
finben: Done!

Hymns/Tags/TagSheetViewModel.swift
skywalkerdude: Well, you are implementing the "equals" function, and even though you aren't using it now, this stands to be confusing in the future. Basically here you are saying that as long as two tags have the same `title`, then they are equal. But semantically that doesn't make sense, because two tags should only be equal if they have the same title AND color.

Hymns/Tags/TagSheetViewModel.swift
skywalkerdude: Yeah, I feel like this would crash because you aren't registering the HymnIdentifier anywhere. I'd recommend you just remove this and instantiate it with the constructor

Hymns/Infrastructure/Data/Models/UiHymn.swift
```diff
@@ -15,6 +16,25 @@ struct UiHymn: Equatable {
...
+    /*
+    func computedTitle(hymn: HymnIdentifier) -> String {
+        let title: Title
+        if hymn.hymnType == .classic {
+            title = "Hymn \(hymn.hymnNumber)"
+         } else {
+            title = self.title.replacingOccurrences(of: "Hymn: ", with: "")
+         }
+         return title
+    } */
```
skywalkerdude, line 28: remove

Hymns/Tags/TagEntity.swift
skywalkerdude: Do these need to be UIColors? I think you can change them to just Color and use `Color(UIColor(...))`

Hymns/Tags/TagSheetViewModel.swift
finben: What do you mean? I can't initiate without hymnIdentifier because it is a required argument. It's not resolved in the constructor.

Hymns/Tags/TagSheetViewModel.swift
finben: Ahhh, thanks good point. Thanks for the explanation. Makes more sense now. Resolved in recent commit

Hymns/Tags/TagSheetViewModel.swift
skywalkerdude: I meant, remove the Resolver extension here, and when you need the view model, just create it manually. `TagSheetViewModel(hymnToDisplay: ...)`

finben: Okay, cool done!

Hymns/Tags/TagSheetViewModel.swift
finben: done

Hymns/Tags/TagEntity.swift
finben: Nice! I don't know why I didn't think of that lol... Great idea. Done!

finben:
Thanks bro for all the quick reviews!

Issue #228, opened 2020-06-01
Certain PDFs aren't loading


skywalkerdude:
1. Search hymn 441
2. Open languages
3. Select the first Chinese song (诗歌(简))
4. Switch to chords/piano/guitar

light mode:
![Simulator Screen Shot - iPhone 8 - 2020-06-01 at 00 57 34](https://user-images.githubusercontent.com/1427524/83388226-efb70580-a3a2-11ea-950b-27033600d379.png)

dark mode:
![Simulator Screen Shot - iPhone 8 - 2020-06-01 at 00 56 35](https://user-images.githubusercontent.com/1427524/83388168-d6ae5480-a3a2-11ea-83f1-3e1149d61e7e.png)

PR #273, opened 2020-06-11
PDF error states

finben:
Addresses #228 I think it looks pretty good. The only question I have is should we be using the same error state for no url as when it doesn't load? On one hand it looks nice that the error state would be consistent. But if there is no url maybe our error state isn't about us saying "something went wrong" as much as it is about there just not being a url for that song because it doesn't have sheet music? 
 
Simulator Screen Shot - iPhone 8 - 2020-06-11 at 13 39 53


skywalkerdude:
That's a good question. I think if the URL is empty, we probably just don't show the tab at all, right? I believe we already take care of that in DisplayHymnViewModel. Can you just change it to a non-optional and see if it works? I believe it should just work off the bat

Hymns/Common/PDFViewer.swift
```diff
@@ -13,7 +13,7 @@ struct PDFViewer: UIViewRepresentable {
 
     func updateUIView(_ pdfView: PDFView, context: Context) {
         guard let url = url else {
-            // TODO create error state
+            displayErrorState(pdfView)
             return
```
skywalkerdude, line 17: I think url can just straight up be non-optional, and everything should still work

finben: The URL type is always an optional value. I don't even think it is possible to store a URL without an optional. Can I just do `guard let url = url else {
return } ` and leave it at that

finben:
Lol... well you proved me wrong xD

PR #274, opened 2020-06-11
Lotte, another attempt. Let's see if it passes tests

finben:


finben:
YAH!! This seems like it is passing tests! Can you review. I implemented it a bit differently this time and avoided using overlay which I think was breaking out tests. The only thing is I had to turn off copywrite header danger file for this to pass because of all the pods in it.

How do you handle putting in new pods with the danger swift copywrite thing. Do we just turn it off merge it in and then turn it back on?

Hymns/Home/HomeContainerView.swift
```diff
@@ -4,34 +4,54 @@ import Resolver
...
-                    .tag(HomeTab.browse)
-                    .hideNavigationBar()
+            if showSplash {
+                return SplashScreenView()
+                    .opacity(showSplash ? 1 : 0)
+                    .onAppear {
+                        DispatchQueue.main.asyncAfter(deadline: .now() + 2.7) {
+                            self.showSplash = false
+                        }
+                }.eraseToAnyView()
```
skywalkerdude, line 19: I think, instead of sticking everything in here, can we just create another view before this, call it LaunchView or something that will either show the splash, or navigate straight to the HomeContainerView. Is that possible?

skywalkerdude:
> How do you handle putting in new pods with the danger swift copywrite thing. Do we just turn it off merge it in and then turn it back on?

I'll use my admin privileges squash this PR for you once we get this good to go. That way we won't need to disable anything. I can just bypass it ;)

skywalkerdude:
> YAH!! This seems like it is passing tests! Can you review. I implemented it a bit differently this time and avoided using overlay which I think was breaking out tests.

Could you elaborate a little bit on this? What overlays are you referring to? And how come it broke the tests?

finben:
Remember when I was trying to implement this before like 4 super random tests would break. We got failed tests like RealmImplSpec accessing Realm from wrong thread. The overlays I am talking about is the way I had gone about implementing this before was using an overlay and opacity would change instead of just displaying different views based on a state variable like I am trying now

finben:
@skywalkerdude As soon as I move it to its own view or route to views it once again fails a bunch of random tests. I think it throws the tests of that it is not at the home container right away maybe?

Hymns/Home/LaunchRouterView.swift
```diff
@@ -0,0 +1,26 @@
...
+            if showSplash {
+                return SplashScreenView()
+                    .opacity(showSplash ? 1 : 0)
+                    .onAppear {
+                        DispatchQueue.main.asyncAfter(deadline: .now() + 2.7) {
+                            self.showSplash = false
+                        }
+                }.eraseToAnyView()
+            } else {
+                return HomeContainerView().eraseToAnyView()
```
skywalkerdude: So here you are showing the HomeContainer inside of the LaunchRouterView. This means that all future views will be nested inside of this LauncherView.

Is it not possible in SwiftUI to push and pop views programmatically?

Hymns/Home/HomeContainerView.swift
skywalkerdude: I came across this: https://stackoverflow.com/a/62225392/1907538. Do you think something like that is feasible?

skywalkerdude: another possibility: https://stackoverflow.com/a/57274613/1907538

Since we aren't using the back button in this case, maybe we can get away with it?

Hymns/Home/LaunchRouterView.swift
skywalkerdude: oops, I meant to post my comments here: 

> I came across this: https://stackoverflow.com/a/62225392/1907538. Do you think something like that is feasible?

> another possibility: https://stackoverflow.com/a/57274613/1907538
> Since we aren't using the back button in this case, maybe we can get away with it?

finben:
Can you take a look at this. The only thing failing is the danger copywrite header. I got tests passing, I went with an implementation that basically just circumvents testing for Debug. That's because I think the tests are just being thrown off by the animation view. So we just swap the root view for our Debug versions and they go straight to home view.

Hymns/Home/LaunchRouterView.swift
finben: I don't think that is really a problem because we don't declare a navigationView in launch router. So the root for our navigation stack should still be HomeContainer right?

finben:
@skywalkerdude Can I get an update on this PR? It should be good. Passing everything but Danger because of the copywrite

skywalkerdude:
ohhh sorry. I didn't realize you were waiting on me. I figured this was lower priority than some of the other stuff we were working on, so I was prioritizing those instead. I'll try to take a look into this later tonight?

(But in the meantime, can you focus on the bugs that Emily filed?)

finben:
Yeah, will do. I've been taking it easy today and probably will just do a
little this weekend. But I'm planning to be doing alot of Swift next week
with WWDC

Luke <notifications@github.com> schrieb am Fr., 19. Juni 2020, 9:43 PM:

PR #275, opened 2020-06-12
Tag front end

finben:
Part 2 lol...


Hymns/Common/CustomColors.swift
```diff
@@ -0,0 +1,14 @@
...
+    static let backgroundBlue = Color(red: 2/255, green: 118/255, blue: 254/255, opacity: 0.2)
+    static let backgroundGreen = Color(red: 80/255, green: 227/255, blue: 194/255, opacity: 0.2)
+    static let backgroundYellow = Color(red: 255/255, green: 209/255, blue: 0/255, opacity: 0.2)
+    static let backgroundRed = Color(red: 255/255, green: 0, blue: 31/255, opacity: 0.2)
+
+    static let foregroundBlue = Color(red: 2/255, green: 118/255, blue: 254/255, opacity: 1.0)
+    static let foregroundGreen = Color(red: 35/255, green: 190/255, blue: 155/255, opacity: 1.0)
+    static let foregroundYellow = Color(red: 176/255, green: 146/255, blue: 7/255, opacity: 1.0)
+    static let foregroundRed = Color(red: 255/255, green: 0, blue: 31/255, opacity: 0.78)
+}
```
skywalkerdude, line 14: I don't think this would be used outside of tags right? We should make these part of the TagColors enum. Just create a `backgroundColor` param and a `foregroundColor` param kind of like how I did the with `abbreviatedValue` in HymnType

Hymns/Display/DisplayHymnBottomBar.swift
```diff
@@ -45,6 +46,12 @@ struct DisplayHymnBottomBar: View {
                     }
                 }
             }
+            if tagOn {
+                VStack {
+                    Divider()
+                    TagSheetView(viewModel: TagSheetViewModel(hymnToDisplay: self.viewModel.identifier), tagOn: self.$tagOn).eraseToAnyView()
+                }
+            }
```
skywalkerdude, line 54: Can we make this a bottomSheet instead?

Hymns/Display/DisplayHymnBottomBar.swift
```diff
@@ -204,7 +213,7 @@ struct DisplayHymnBottomBar_Previews: PreviewProvider {
         var dialogBuilder: (() -> AnyView)?
         return DisplayHymnBottomBar(dialogBuilder: Binding<(() -> AnyView)?>(
             get: {dialogBuilder},
-            set: {dialogBuilder = $0}), viewModel: viewModel).toPreviews()
+            set: {dialogBuilder = $0}), tagOn: .constant(true), viewModel: viewModel).toPreviews()
```
skywalkerdude, line 216: woah. I didn't know you could do that!

Hymns/Tags/TagEntity.swift
```diff
@@ -69,13 +69,13 @@ extension TagColor {
...
+            return CustomColors.foregroundBlue
         case .green:
-            return Color(red: 35/255, green: 190/255, blue: 155/255, opacity: 1.0)
+            return CustomColors.foregroundGreen
         case .yellow:
-            return Color(red: 176/255, green: 146/255, blue: 7/255, opacity: 1.0)
+            return CustomColors.foregroundYellow
         case .red:
-            return Color(red: 255/255, green: 0, blue: 31/255, opacity: 0.78)
+            return CustomColors.foregroundRed
```
skywalkerdude, line 78: what's wrong with just putting the colors directly in here?

Hymns/Common/CustomColors.swift
finben: The tricky part is I also use them in TagSelectedLabelView. But in that case I need to be able to use them as a static since there is not an instantiated TagColor that is being switched on. But I also use these same colors as an instance property for TagColor so in that case they are not static. 

I did something similiar in my latest commit to what you did in HymnType. Lmk what you think. The problem with it is that it seemed to still add alot more code. Additionally, I can't use the enum function call within the stroke modifer in SelectLabelView so have to now hardcode the color there.  Also it introduced a wierd issue where the color is not being displayed correctly for SelectLabelView even though nothing changed for the colors. 
Simulator Screen Shot - iPhone 8 - 2020-06-12 at 09 43 20

Hymns/Tags/TagEntity.swift
finben: It's just because I use them again in TagSelectedLabelView and there they are needed without TagColor being instantiated. I did do this in my recent commit. But it introduces several other problems as I mentioned above and I don't know if it is a commit worth keeping.

Hymns/Display/DisplayHymnBottomBar.swift
finben: So, I had done that previously in some earlier commits. But it was not resizing correctly whenever the tags would go to empty or populate as is needed from the figma specs. When the view would update and redraw it would be buggy with the tags sometimes not redrawing onto the view. This way here is super smooth. It redraws to the correct size everytime and is still coming up at the bottom. 

What is the benefit of making it a bottomSheet?

skywalkerdude: Can you elaborate what you mean? Or show me a video? I just tried it and it seems okay. The only problem is that I don't know how to make the sheet non-full screen

I think the benefit is
1) We don't need to do the opacity and stuff ourselves
2) There's a nice little animation that also minimizes the background a little bit
3) It's generally better not to reinvent the wheel if we can use something built-in

Hymns/Common/CustomColors.swift
skywalkerdude: I don't see anything called TagSelectedLabelView

Hymns/Tags/SelectLabelView.swift
```diff
@@ -0,0 +1,107 @@
...
+        self.redButton = false
+        self.yellowButton.toggle()
+    }
+
+    private func resetForRed() {
+        self.blueButton = false
+        self.greenButton = false
+        self.yellowButton = false
+        self.redButton.toggle()
+    }
```
skywalkerdude: I'm sure you can think of a better way to do this lol. How about you pass in the button that is true and you set everything else to false?

Hymns/Tags/SelectLabelView.swift
```diff
@@ -0,0 +1,107 @@
+import SwiftUI
+
+struct SelectLabelView: View {
+    @Binding var tagColor: TagColor
+    @State var blueButton = false
+    @State var greenButton = false
+    @State var yellowButton = false
+    @State var redButton = false
```
skywalkerdude: what if you did this:

`@State colorSelected: TagColor = .none`

Then when you set the backgrounds:

`.stroke(Color.blue, lineWidth: self.colorSelected == .blue ? 3 : 0)`

Hymns/Tags/SelectLabelView.swift
```diff
@@ -0,0 +1,107 @@
+import SwiftUI
+
+struct SelectLabelView: View {
```
skywalkerdude: `ColorSelectorView` makes more sense

Hymns/Tags/TagSheetView.swift
```diff
@@ -0,0 +1,62 @@
...
+            TextField("Label it however you like", text: self.$tagName)
+            Divider()
+            ColorSelectorView(tagColor: self.$tagColor).padding(.vertical)
+            if !self.viewModel.tags.isEmpty {
+                Section {
+                    Text("Tags").font(.body).fontWeight(.bold)
+                    HStack {
+                        ForEach(self.viewModel.tags, id: \.self) { tag in
+                            Button(action: {
+                                //TODO FIX Delete is deleting all the tags because the button when clicked is clicking all of them.
```
skywalkerdude, line 29: Why does this delete all the tags? Is it difficult to just fix that in this PR

Hymns/Tags/TagSheetView.swift
```diff
@@ -0,0 +1,62 @@
...
+                                HStack {
+                                    Text(tag.title).font(.body).fontWeight(.bold)
+                                    Image(systemName: "xmark.circle")
+                                }
+                            })
+                                .padding(10)
+                                .foregroundColor(tag.color.foreground)
+                                .background(tag.color.background)
+                                .cornerRadius(20)
+                                .lineLimit(1)
```
skywalkerdude, line 41: Why line limit 1? What happens if someone enters a really long tag?

Hymns/Tags/TagEntity.swift
```diff
@@ -72,10 +76,46 @@ extension TagColor {
...
             return Color(red: 35/255, green: 190/255, blue: 155/255, opacity: 1.0)
+
         case .yellow:
             return Color(red: 176/255, green: 146/255, blue: 7/255, opacity: 1.0)
         case .red:
             return Color(red: 255/255, green: 0, blue: 31/255, opacity: 0.78)
         }
     }
+
+    var abbreviatedValue: String {
```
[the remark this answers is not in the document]
skywalkerdude: fixed for you. feel free to resolve if it makes sense.

Hymns/Tags/SelectLabelView.swift
skywalkerdude: fixed for you. feel free to resolve if it makes sense.

Hymns/Tags/SelectLabelView.swift
skywalkerdude: fixed for you. feel free to resolve if it makes sense.

Hymns/Tags/SelectLabelView.swift
skywalkerdude: fixed for you. feel free to resolve if it makes sense.

Hymns/Display/DisplayHymnBottomBar.swift
finben: So the way I had it automatically redraws when a tag is removed or added. See the two images below and how the size changes to match what is necessary. What I mean't though about not redrawing and stuff was when I was using a custom sized geometry reader bottom view so as to make it not take up the whole screen. I think the whole solution I was trying to develop is with the point that we don't want to use a whole sheet that covers everything. 

I think the main thing is that everything I have been doing here is basically with the idea that we need a solution that is not full screen like built in sheet. And unfortunately, sheet at this point just does not support anything but being full size. 

1. I agree it is annoying to have to do opacity ourselves but unfortunately that is going to be necessary if we don't want it to take the whole screen with a big sheet. 

To me it's worth the custom implementation in order to gain the result of it not being a full sheet and matching the figma sheet better. But I do wish the backgound animation and blur can be improved. Idk what do you think? Go with the full sheet?

Simulator Screen Shot - iPhone 8 - 2020-06-13 at 09 44 42

Simulator Screen Shot - iPhone 8 - 2020-06-13 at 09 44 48

finben: We can add some blur as well as opacity to make it look more like a sheet though
Simulator Screen Shot - iPhone 8 - 2020-06-13 at 10 11 57
. See below

Hymns/Tags/SelectLabelView.swift
finben: I changed the stroke color from always being blue. But other than that this is great!

Hymns/Tags/SelectLabelView.swift
finben: Lol, yeah nice!

Hymns/Tags/TagSheetView.swift
finben: It's a wierd bug thing. It's a problem with the front end. I think it has to do with there being a bunch of buttons within a ForEach and how it SwiftUI reacts when one of them is clicked. Somehow they are all getting triggered sometimes. But also sometimes it doesn't happen. I need to figure out a solution for this. Because it is a significant problem.

finben: Nevermind. I found a solution. You just can't use button.                             //https://stackoverflow.com/questions/60960077/how-do-i-keep-multiple-button-actions-separate-in-swiftui-foreach-content Can't use a button here
I implemented the solution using onTapGesture. I left a comment in the code but maybe we don't need a code comment idk?

Hymns/Tags/TagSheetView.swift
finben: It applies to the tag. If it's a long tag it just gets .... 
The reason is otherwise it looks pretty janky. Here is it without linelimit 1 
Simulator Screen Shot - iPhone 8 - 2020-06-13 at 10 51 59

Hymns/Tags/SelectLabelView.swift
finben: Looks good!

Hymns/Tags/TagSheetView.swift
```diff
@@ -25,20 +25,19 @@ struct TagSheetView: View {
...
-                            }, label: {
-                                HStack {
-                                    Text(tag.title).font(.body).fontWeight(.bold)
-                                    Image(systemName: "xmark.circle")
+
+                            //https://stackoverflow.com/questions/60960077/how-do-i-keep-multiple-button-actions-separate-in-swiftui-foreach-content Can't use a button here
+                            HStack {
+                                Text(tag.title).font(.body).fontWeight(.bold)
+                                Image(systemName: "xmark.circle").onTapGesture {
+                                    self.viewModel.deleteTag(tagTitle: tag.title, tagColor: .blue)
```
skywalkerdude, line 33: tag.color

finben:
So I added back the VStack instead of sheet. It still is not generic. If you have any ideas to make it more generic please help me lol... But I did move from using an opacity change to doing a blur which looks alot better especially in dark mode. 
Simulator Screen Shot - iPhone 8 - 2020-06-13 at 11 55 55

finben:
Also, note in the screenshot that right now tags can be saved without color. I am guessing we don't want that right? Or can a user save a tag without a color?

skywalkerdude:
> Also, note in the screenshot that right now tags can be saved without color. I am guessing we don't want that right? Or can a user save a tag without a color?

I think it's fine too save without color

skywalkerdude:
> So I added back the VStack instead of sheet. It still is not generic. If you have any ideas to make it more generic please help me lol... But I did move from using an opacity change to doing a blur which looks alot better especially in dark mode.
> Simulator Screen Shot - iPhone 8 - 2020-06-13 at 11 55 55

Okay. I'll work on this. Can you take a look at some of the other blockers?

Issue #174, opened 2020-05-22
Hide "song info" if there is no info to show


skywalkerdude:
Rare occasion, but still technically feasible:
![Simulator Screen Shot - iPhone 8 - 2020-05-22 at 00 28 20](https://user-images.githubusercontent.com/1427524/82642698-2df23f00-9bc3-11ea-8dd3-ba958e071626.png)

What would be a good experience to show the empty state?

Emyeh:
@skywalkerdude If there's no song info we should just disable to the info icon and making at 25% opacity.

Emyeh:
Actually since the buttons within the toolbar are dynamic we should just remove the info icon or option completely if there is none.

skywalkerdude:
Repurposing task for removing the icon/option if there is no song info to show

Issue #173, opened 2020-05-22
Add composer, key, time, meter, scriptures, hymnCode to the song info dialog


skywalkerdude:
See https://github.com/HymnalDreamTeam/Hymns-iOS/pull/172 as a reference

PR #277, opened 2020-06-12
close 2 issues

skywalkerdude:
closes https://github.com/HymnalDreamTeam/Hymns-iOS/issues/173
closes https://github.com/HymnalDreamTeam/Hymns-iOS/issues/174


Hymns/Display/DisplayHymnBottomBar.swift
```diff
@@ -175,6 +177,8 @@ struct DisplayHymnBottomBar: View {
     }
 }
 
+
+
```
finben, line 181: Excess space? remove?

Hymns/Display/Song Info/SongInfoDialogViewModel.swift
```diff
@@ -23,8 +23,10 @@ class SongInfoDialogViewModel: ObservableObject {
         self.identifier = identifier
         self.mainQueue = mainQueue
         self.repository = repository
+        fetchSongInfo()
```
finben, line 26: This is a good idea. Should we be doing this on some of our other view models instead of using the view with on appear?

skywalkerdude: No. I only did this because we need to know ahead of time whether or not the song info list is populated or not so we know whether or not to show the song info button. But in general, we shouldn't make the request until onAppear.

PR #278, opened 2020-06-12
Adding Hymn number to classic songs in search

rachely3n:
When classic hymns are searched, include hymn with number. See screenshot below:

Screen Shot 2020-06-12 at 8 44 25 AM


skywalkerdude:
>.< I think you misunderstood again. We're not trying to add `Hymn: ` to places that are already displaying the title, but we are trying to add ` : {TITLE}` to places that are only displaying the hymn number.

So for example, the recent songs list: 
Simulator Screen Shot - iPhone 8 - 2020-06-12 at 12 23 22

Or the favorited songs list:
Simulator Screen Shot - iPhone 8 - 2020-06-12 at 12 23 38

We need to show the title here, but we actually don't want to show the `Hymn: ` in the search results because it makes things kind of cluttered.

Does that make sense?

rachely3n:
O didn't realize those places were only doing that... gotcha. At least I'm familiarizing myself with the code base rip. Let me close this one out.

PR #284, opened 2020-06-13
Make about us link to external site and fix wording

finben:


Hymns/Settings/AboutUsInfoView.swift
```diff
@@ -3,47 +3,45 @@ import UIKit
...
-                    //https://stackoverflow.com/questions/59359730/is-it-possible-to-add-an-in-line-button-within-a-text
+            VStack {
+                HStack {
+                    Button(action: {
+                        self.presentationMode.wrappedValue.dismiss()
+                    }, label: {
+                        Image(systemName: "xmark")
+                    })
+                    Text("About us").fontWeight(.bold).padding(.leading)
+                    Spacer()
```
skywalkerdude, line 18: FYI, If you make the VStack's alignment .leading, then you don't need these HStack-Spacer combos

finben: Nice!!

Issue #283, opened 2020-06-13
icons don't look evenly spaced


skywalkerdude:
![Simulator Screen Shot - iPhone 8 - 2020-06-12 at 23 11 33](https://user-images.githubusercontent.com/1427524/84561547-23c6ea80-ad02-11ea-94ef-ffa133f8f41a.png)

PR #285, opened 2020-06-13
Fix tag icon spacing issue

finben:
Deals with #283

### DIFF
--- a/Hymns/Display/DisplayHymnBottomBar.swift
+++ b/Hymns/Display/DisplayHymnBottomBar.swift
@@ -106,7 +106,9 @@ struct DisplayHymnBottomBar: View {
                         Button(action: {
                             self.showAudioPlayer.toggle()
                         }, label: {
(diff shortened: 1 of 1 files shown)


Hymns/Display/DisplayHymnBottomBar.swift
```diff
@@ -106,7 +106,7 @@ struct DisplayHymnBottomBar: View {
                         Button(action: {
                             self.showAudioPlayer.toggle()
                         }, label: {
-                            showAudioPlayer ? Image(systemName: "play.fill").accentColor(.accentColor) : Image(systemName: "play").accentColor(.primary)
+                            (showAudioPlayer ? Image(systemName: "play.fill").accentColor(.accentColor) : Image(systemName: "play").accentColor(.primary)).padding()
```
skywalkerdude, line 109: I think the problem isn't that it needs a `padding()`, but that it needs to be a `BottomBarLabel` not an `Image`

finben: Looks good. I didn't know I could put a view in a conditional modifier like that! sweet!

PR #287, opened 2020-06-14
Tag front end2

skywalkerdude:
Ben's code using vstack


finben:
I think we are done with this no? I am going to close it down

Issue #288, opened 2020-06-14
Add creation date to tags db and sort by creation time...


skywalkerdude:

skywalkerdude:
blocked by https://github.com/HymnalDreamTeam/Hymns-iOS/pull/275

PR #289, opened 2020-06-14
add a created date field to TagEntity so we can order then when we get them out of the db

skywalkerdude:


skywalkerdude:
fixes https://github.com/HymnalDreamTeam/Hymns-iOS/issues/288

PR #297, opened 2020-06-14
Small text resizing patch

finben:


Hymns/Display/VerseView.swift
```diff
@@ -8,15 +8,15 @@ struct VerseLineView: View {
 
     var body: some View {
         HStack(alignment: .top) {
-            Text(viewModel.verseNumber ?? "").font(viewModel.fontSize.font).frame(minWidth: viewModel.fontSize.minWidth)
+            Text(viewModel.verseNumber ?? "").font(viewModel.fontSize.font).frame(minWidth: 12)
```
finben, line 11: We kind of need a definite size here. This causes the wierd issue especially when you go from big to small text of extra space.

Hymns/Display/VerseView.swift
```diff
@@ -8,15 +8,15 @@ struct VerseLineView: View {
...
             VStack(alignment: .leading) {
                 if transliterate {
                     viewModel.transliteration.map { transliteration in
                         Text(transliteration).font(viewModel.fontSize.font)
                     }
                 }
                 Text(viewModel.verseText).font(viewModel.fontSize.font)
-            }.lineSpacing(5).padding(.bottom, 5)
+                }.fixedSize(horizontal: false, vertical: true).padding(.bottom, 5).lineSpacing(5)
```
finben, line 19: This deals with the issue of having lyrics do .... and us losing words on songs. This eliminates the issue entirely and just has it wrap to the next line.

Hymns/Display/VerseViewModel.swift
```diff
@@ -22,9 +22,7 @@ class VerseLineViewModel: Hashable, ObservableObject {
         userDefaultsManager
             .fontSizeSubject
             .sink { fontSize in
-                withAnimation(.easeInOut(duration: 0.5)) {
-                    self.fontSize = fontSize
-                }
+                self.fontSize = fontSize
         }.store(in: &disposables)
```
finben, line 26: I could not find a single animation that results in us not having the ... during transition. To me it's probably better to just forgo the animation all together than have it feed in the odd manner it is now.

skywalkerdude:
This doesn't fix the issue where the text isn't aligned correctly after resizing from large/extra large to normal:
Simulator Screen Shot - iPhone 8 - 2020-06-14 at 17 02 38 vs 
Simulator Screen Shot - iPhone 8 - 2020-06-14 at 17 02 33

Hymns/Display/VerseViewModel.swift
skywalkerdude: agreed

Hymns/Display/VerseView.swift
skywalkerdude: Without the animation, the "..." never appears anymore, so I think we can get rid of this

Hymns/Display/VerseView.swift
skywalkerdude: I still get the extra space: Simulator Screen Shot - iPhone 8 - 2020-06-14 at 17 02 33

Hymns/Display/VerseView.swift
finben: It happens on animation but it also happens when a lines text does not fit. For instance try an iphone 8 emulation and the song "in this godless age."

finben:
Okay, I got a fix... Just don't ask me how it works... lol. Also, take a look at this thing I found to trim the whitespaces. I feel like its nice one hymns like 498 with janky whitespace coming from hymnal.net. But you may not agree.

Hymns/Display/HymnLyricsView.swift
```diff
@@ -36,7 +36,7 @@ public struct HymnLyricsView: View {
                         ForEach(lyrics, id: \.self) { verseViewModel in
                             VerseView(viewModel: verseViewModel, transliterate: self.$transliterate)
                         }
-                    }.padding()
+                    }.frame(minWidth: 0, maxWidth: .infinity, alignment: .leading).padding()
```
skywalkerdude, line 39: I think this is the key line. Everything else, I removed it, and it seemed to still work

Hymns/Display/VerseView.swift
skywalkerdude: I think we can revert this right?

Hymns/Display/VerseView.swift
skywalkerdude: ah, gotcha. you are right!

Hymns/Display/VerseView.swift
```diff
@@ -8,15 +8,17 @@ struct VerseLineView: View {
...
             VStack(alignment: .leading) {
                 if transliterate {
                     viewModel.transliteration.map { transliteration in
-                        Text(transliteration).font(viewModel.fontSize.font)
+                        Text(transliteration.trimmingCharacters(in: .whitespacesAndNewlines)).font(viewModel.fontSize.font)
                     }
                 }
-                Text(viewModel.verseText).font(viewModel.fontSize.font)
-                }.fixedSize(horizontal: false, vertical: true).padding(.bottom, 5).lineSpacing(5)
+                Text(viewModel.verseText.trimmingCharacters(in: .whitespacesAndNewlines)).font(viewModel.fontSize.font)
```
skywalkerdude: I think the whitespace is okay. It makes sense to indent some of the lines (even in the hymnal they're indented), so I would revert this part.

Hymns/Display/HymnLyricsView.swift
```diff
@@ -13,11 +13,11 @@ public struct HymnLyricsView: View {
             guard let lyrics = viewModel.lyrics else {
                 return Text("Lyrics are not available for this song").maxSize().eraseToAnyView()
             }
-
+
```
skywalkerdude: swiftlint

Hymns/Display/HymnLyricsView.swift
```diff
@@ -13,11 +13,11 @@ public struct HymnLyricsView: View {
             guard let lyrics = viewModel.lyrics else {
                 return Text("Lyrics are not available for this song").maxSize().eraseToAnyView()
             }
-
+            
             guard !lyrics.isEmpty else {
                 return ActivityIndicator().maxSize().eraseToAnyView()
             }
-
+
```
skywalkerdude: swiftlint

Hymns/Display/HymnLyricsView.swift
finben: Yeah, that is the line.

Hymns/Display/VerseView.swift
finben: reverted in recent commit

Hymns/Display/HymnLyricsView.swift
finben: addressed

Issue #291, opened 2020-06-14
'Cancel' and 'Add' button should be below the tags created


Emyeh:
![IMG_7557](https://user-images.githubusercontent.com/26554065/84587279-a3bc8580-add2-11ea-8d1e-c66933bc7aba.PNG)

Also thinking after a user adds a tab should we change the 'cancel' button to 'done'. Or place 'add' somewhere else. It might be confusing since 'cancel' doesn't get rid of the tags added. I'll need to revisit the CTAs here.

skywalkerdude:
Hmm, it doesn't seem to resize correctly either. It seems to look like this after resizing from extra large back to normal: 
![Simulator Screen Shot - iPhone 8 - 2020-06-14 at 13 48 26](https://user-images.githubusercontent.com/1427524/84603767-c340c600-ae45-11ea-9708-5e98d9dca52d.png)

skywalkerdude:
@Emyeh is it a big deal to you that the buttons are above the tag list instead of below? There's a limitation with the iOS UI framework we are using that causes the add/cancel button to be docked at the very bottom of the screen, so there's a huge gap in the middle. I was hoping to ship with it looking like this for now and hopefully at WWDC 20 they'll give us some solutions.

finben:
@skywalkerdude  😂 Lol! I just took a look at this. Your geometry reader taking up to much space xD

Emyeh:
Forget what I said about changing the button from 'Cancel' to 'Done' after adding a tag. It'll be clearer just to rename 'Cancel' to 'Close'.

PR #298, opened 2020-06-15
hehe it works. Passing a whole view like a view model interesting

finben:
Addresses #291 

Simulator Screen Shot - iPhone 8 - 2020-06-14 at 22 49 20


skywalkerdude:
Nice!!! Great find!

I took your idea and refactored it so that the tag logic isn't inside the WrappedHStack (thus keeping it generic), and it seems to be working pretty well. Please patch this and ensure that it is working as expected! https://github.com/HymnalDreamTeam/Hymns-iOS/pull/299

finben:
I am just going to cancel this and we can push in your #PR 299

Issue #307, opened 2020-06-15
ios 13.3 still has that big gap


skywalkerdude:
To install ios 13.3 simulator: https://stackoverflow.com/questions/58072572/how-to-run-a-different-version-of-the-ios-simulator-on-xcode-11

<img width="1355" alt="Screen Shot 2020-06-15 at 2 33 26 PM" src="https://user-images.githubusercontent.com/1427524/84708362-349b7a00-af15-11ea-876a-91b817d82851.png">

finben:
Bro, this issue is bad news bears.... Basically, it has to do with our NavigationView containing the TabView within our homeContainer. The problem is if we put individual navigationViews within each of the tabs as many users would suggest then we are going to be showing the Tabs on every single page of our app. As of now this is not yet supported in SwiftUI.  https://stackoverflow.com/questions/58067390/hide-tabview-after-clicking-on-a-navigationlink-in-swiftui

I am not sure what to do. We can hope that WWDC 2020 gives us a .hidetabview option which they might do similiar to what they have for navTitle. 

Another, potential solution would be we fake the tabView. And instead of a real tab view we use Buttons in an HStack that essentially just do NavigationBarLink. 

In the mean time I am going to submit a PR that at least shrinks that extra whitespace some.

skywalkerdude:
@finben I'm a little confused. I don't understand that that link is saying. It talks about hiding the tab. But it doesn't talk about like the big gap at the top. Why is it that the NavigationView causes the gap?

finben:
It seems like a bug to me that got resolved in later versions which occurs whenever you place a TabView within a navigationView. It is because at this point Apple only really planned for people to be using navigationView within TabView instead of the other way around.

Issue #135, opened 2020-05-11
Landscape fix on hymn page


Emyeh:
Everything looks really good on landscape except the hymn page:
1. It's too tight at the top, making it hard to tap on the left carat and favorite buttons. We need about a 16dp padding there.
2. The stanzas just need to be centered within the page (they can stay left aligned).

![IMG_7412](https://user-images.githubusercontent.com/26554065/81532243-ee9d3600-9318-11ea-9e78-f35f3a7f02c1.PNG)

@finben @skywalkerdude

finben:
@Emyeh For the lyrics being centered. Is that only in landscape mode that they should be centered or does that also apply to vertical?

Emyeh:
Let's apply it for both so we can have some consistency between the two views.

PR #308, opened 2020-06-15
Decreases the excess space.

finben:
Partially addresses #307  At least cuts that whitespace in about half
Simulator Screen Shot - iPhone 8 - 2020-06-15 at 18 35 50

### DIFF
--- a/Hymns/Display/DisplayHymnView.swift
+++ b/Hymns/Display/DisplayHymnView.swift
@@ -17,7 +17,7 @@ struct DisplayHymnView: View {
                 ActivityIndicator().maxSize()
             } else {
                 VStack(spacing: 0) {
-                    DisplayHymnToolbar(viewModel: viewModel)
+                    DisplayHymnToolbar(viewModel: viewModel).padding(.top, 5)
                     if viewModel.tabItems.count > 1 {
                         IndicatorTabView(currentTab: $viewModel.currentTab, tabItems: viewModel.tabItems)
                     } else {
--- a/Hymns/Display/DisplayHymnView.swift
+++ b/Hymns/Display/DisplayHymnView.swift
@@ -17,7 +17,7 @@ struct DisplayHymnView: View {
                 ActivityIndicator().maxSize()
             } else {
                 VStack(spacing: 0) {
-                    DisplayHymnToolbar(viewModel: viewModel).padding(.top, 5)
+                    DisplayHymnToolbar(viewModel: viewModel)
                     if viewModel.tabItems.count > 1 {
                         IndicatorTabView(currentTab: $viewModel.currentTab, tabItems: viewModel.tabItems)
                     } else {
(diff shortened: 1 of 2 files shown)


finben:
The second commit also addresses landscape mode issues like the following. And the other mentioned by Emily in #135
Simulator Screen Shot - iPhone 8 - 2020-06-15 at 18 53 11

Hymns/Display/DisplayHymnView.swift
```diff
@@ -17,7 +17,7 @@ struct DisplayHymnView: View {
                 ActivityIndicator().maxSize()
             } else {
                 VStack(spacing: 0) {
-                    DisplayHymnToolbar(viewModel: viewModel)
+                    DisplayHymnToolbar(viewModel: viewModel).padding(.top, 5)
```
skywalkerdude, line 20: What's this toolbar about?

Hymns/Home/HomeContainerView.swift
```diff
@@ -27,10 +28,13 @@ struct HomeContainerView: View {
                     .tabItem {HomeTab.settings.getImage(selectedTab == HomeTab.settings).imageScale(.large)}
                     .tag(HomeTab.settings)
                     .hideNavigationBar()
-            }.onAppear {
+            }.hideNavigationBar().padding(.top, self.iosAbove134 ? 5 : 0).onAppear {
```
skywalkerdude, line 31: I'm confused here. So instead of removing the top padding for ios below 13.4, you're adding padding if it is ios 13.4 and above??

Hymns/Display/DisplayHymnView.swift
skywalkerdude, line 20: I think it already looks fine without the padding. Emily's bug was an old bug and I have since adding padding to the toolbar's components
Simulator Screen Shot - iPhone 8 - 2020-06-15 at 19 08 08

Hymns/Home/HomeContainerView.swift
finben: Removing the top padding for ios below 13.4 does nothing. The hideNavigationBar helps though. The thing is I wanted to add padding for the iOS above 13.4 to help us in landscape if you see my screenshot. But that will then add more padding to iOS 13.3 which we obviously want to avoid since it has too much as is.

skywalkerdude: I see. Could you take some before and after screenshots? I tried to remove the padding, and it looks fine to me...

skywalkerdude: without padding:
Screen Shot 2020-06-15 at 9 50 38 PM
Screen Shot 2020-06-15 at 9 50 50 PM
Screen Shot 2020-06-15 at 9 51 04 PM

skywalkerdude: with padding:
Screen Shot 2020-06-15 at 9 52 46 PM
Screen Shot 2020-06-15 at 9 52 56 PM
Screen Shot 2020-06-15 at 9 53 03 PM

finben: Lol when I did this using the same simulator and no padding I get the following. 
Simulator Screen Shot - iPhone 8 - 2020-06-16 at 21 34 45
Simulator Screen Shot - iPhone 8 - 2020-06-15 at 18 54 02

skywalkerdude: lol. Alright well let's just get this on with only the navigation bar, and then I'll create another PR with your changes if I run into it again. That way we can keep the PRs distinct anyway. makes it easier to search through in the future

PR #318, opened 2020-06-16
Chinese search

finben:
This is not a real solution but shows you the work I have so far. As you will see the binding works great when passed into the UIViewRepresentable struct from a SwiftUI State var.  

However, it does not work when we pass in the @published var.

### DIFF
--- a/Hymns/Common/SearchBar.swift
+++ b/Hymns/Common/SearchBar.swift
@@ -4,20 +4,39 @@ import SwiftUI
  * Custom search bar that will animate in a `Cancel` button when activated/selected.
  * https://stackoverflow.com/questions/56490963/how-to-display-a-search-bar-with-swiftui
  */
+
+struct SearchButton: View {
+    @Binding var text: String
+
+    var body: some View {
+
+    Button(action: {
+        self.text = ""
+    }, label: {
+        Image(systemName: "xmark.circle.fill")
+    })
+}
+}
 struct SearchBar: View {
 
     @Binding var searchText: String
     @Binding var searchActive: Bool
+    @State var UISearchText: String = "" //This is the state var that will get passed in to the uiViewRep
     let placeholderText: String
-    @State var titleFocused = false
+
     var body: some View {
-        HStack {
+        if searchText != UISearchText {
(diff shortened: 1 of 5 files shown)


Hymns/Common/SearchBar.swift
```diff
@@ -4,20 +4,39 @@ import SwiftUI
...
+            self.searchText = UISearchText
+        }
+
+        return HStack {
             HStack {
-                Image(systemName: "magnifyingglass").padding(.leading, 6)
-                TextField(placeholderText, text: $searchText)
+                Image(systemName: "magnifyingglass").padding(.leading, 6).onTapGesture {
+                }
+                CustomUIKitTextField(text: self.$UISearchText, placeholder: self.placeholderText)
```
finben, line 36: This is the annoying part. Whenever, I would pass in self.$SearchText it would just never update. However, when I pass in a State var like self.$UISearchText then it updates just fine.

skywalkerdude: Yeah mine doesn't work either. It freezes my simulator. Have you tried putting logs in HomeViewModel?

So like:

```
        $searchActive
            .receive(on: mainQueue)
            .sink { searchActive in
                self.analytics.logSearchActive(isActive: searchActive)
                **PRINT STATEMENT**
                if !searchActive {
                    self.resetState()
                    self.fetchRecentSongs()
                    return
                }
        }.store(in: &disposables)

        $searchParameter
            // Ignore the first call with an empty string since it's take care of already by $searchActive
            .dropFirst()
            // Debounce works by waiting a bit until the user stops typing and before sending a value
            .debounce(for: .seconds(0.3), scheduler: mainQueue)
            .sink { searchParameter in
                **PRINT STATEMENT**
                self.analytics.logQueryChanged(queryText: searchParameter)
                self.refreshSearchResults()
        }.store(in: &disposables)
```

and see what is not being updated. Is it the `searchActive` variable that's not being updated or the `searchParameter` field?

finben: It is the searchParameter. I've debugged and the SearchParameter just always stays empty "" when you pass it in as a binding.

Issue #311, opened 2020-06-16
Dynamic type bug: Tool bar buttons go off screen


Emyeh:
![IMG_7570](https://user-images.githubusercontent.com/26554065/84724459-1d708280-af3d-11ea-8efb-5bc08812cfa9.PNG)

We can keep these buttons the same size as font scales up.

finben:
Issue has been addressed.

PR #325, opened 2020-06-17
put bottom bar into a scroll view

skywalkerdude:
fixes https://github.com/HymnalDreamTeam/Hymns-iOS/issues/311


skywalkerdude:
update: this doesn't work :(

PR #332, opened 2020-06-20
Change search bar hint text to say "Search by number or title"

skywalkerdude:


skywalkerdude:
yup, i replaced them with the correct ones!

Issue #310, opened 2020-06-16
Dynamic type bug: Song info is truncated


Emyeh:
![IMG_7571](https://user-images.githubusercontent.com/26554065/84723200-e77dcf00-af39-11ea-8745-d489594f097e.PNG)

Typically when font size is enlarged, unless it was truncated to begin with it shouldn't truncate. Right now I can't read much about the info because of it. It seems like this happens when the titles and info are within its own sections side by side. 

Here an example of what I'm suggesting (added on Figma): 
![image](https://user-images.githubusercontent.com/26554065/84724258-928f8800-af3c-11ea-8aa1-61b24c5e1332.png)

We can separate the titles vertically so it can wrap nicely. The popup would need to be scrollable.

skywalkerdude:
Need an alternate layout here: https://developer.apple.com/videos/play/wwdc2017/245/ (Minute 15)

finben:
Addressed

PR #356, opened 2020-06-26
Changes songInfo to a sheet, and changes layout for large sizes

finben:
Addresses bug #310

Take note that the solution by the apple Engineer at minute 15 does not work for SwiftUI. However, I found a viable solution here for providing alternate layouts. 
https://stackoverflow.com/questions/58226892/does-swiftui-support-alternate-layouts-by-preferred-font-size

Also, I needed to change dialog to a sheet for large content otherwise there was no way to dismiss since it took the entire view. Also, it looks better. See pictures here of how it looks depending on the size of dynamic text. 

Simulator Screen Shot - iPhone 11 Pro - 2020-06-26 at 08 34 34

Simulator Screen Shot - iPhone 11 Pro - 2020-06-26 at 08 35 58


Hymns/Display/Song Info/SongInfoDialog.swift
```diff
@@ -13,15 +23,28 @@ struct SongInfoDialog: View {
...
+        } else {
+            return VStack(alignment: .leading, spacing: 15) {
+                ForEach(viewModel.songInfo, id: \.self) { songInfo in
+                    SongInfoView(viewModel: songInfo)
+                }
             }
+            .padding()
+            .cornerRadius(5)
+            .background(Color(.secondarySystemBackground))
+            .eraseToAnyView()
```
skywalkerdude, line 46: Instead of doing an alternate layout here, can you just put it in a scrollview regardless of size? I don't think it hurts to be in a scroll view if scrolling is not necessary

Hymns/Display/DisplayHymnBottomBar.swift
```diff
@@ -118,8 +127,12 @@ struct DisplayHymnBottomBar: View {
                 if !viewModel.songInfo.songInfo.isEmpty {
                     Group {
                         Button(action: {
-                            self.dialogBuilder = {
-                                SongInfoDialog(viewModel: self.viewModel.songInfo).eraseToAnyView()
+                            if self.largeSizeCategories.contains(self.sizeCategory) {
```
skywalkerdude, line 130: I saw on the WWDC talk that you can just do `self.sizeCategory >= .extraLarge`. Can you do that instead of defining a list?

Issue #322, opened 2020-06-17
Landscape bug: Center tabs


Emyeh:
![IMG_7583](https://user-images.githubusercontent.com/26554065/84859910-e7a5c980-b022-11ea-9901-82250c49d4fb.PNG)

I think we had it centered at one point but now it remains left aligned in landscape mode.

Previous build:
![image](https://user-images.githubusercontent.com/26554065/84859945-fdb38a00-b022-11ea-91fb-340344d28c93.png)

PR #362, opened 2020-06-29
Provide an alternate layout for the lyrics,chords,guitar,piano bar

finben:
Addresses #322 by using alternate layout based on dynamic size.


Hymns/Common/TabView/TabBar.swift
```diff
@@ -6,39 +6,73 @@ import SwiftUI
...
+                                    key: FirstNonNilPreferenceKey<Anchor<CGRect>>.self,
+                                    value: .bounds,
+                                    transform: { anchor in self.isSelected(tabItem) ? .some(anchor) : nil }
+                            )
+                            Spacer()
+                        }
+                    }
+                }
+            } else {
+                HStack {
```
skywalkerdude, line 46: I don't think this works, because now this is what it looks like without the scroll view:
Simulator Screen Shot - iPhone 8 - 2020-06-29 at 11 49 29

finben:
Okay, I readdressed this by changing the padding to be vertical only. It looks good now on the largest size before the alternate view comes into play.  

Also, note the .minimumscalefactor thing I added. This really hopefully will never do anything. However, in the case that for some reason the text would wrap like in your picture instead it will shrink the text a bit. I think that's a better fail case than the ugly "s" wrapping.

Hymns/Common/TabView/TabBar.swift
```diff
@@ -6,39 +6,73 @@ import SwiftUI
...
+                                value: .bounds,
+                                transform: { anchor in self.isSelected(tabItem) ? .some(anchor) : nil }
+                        )
+                    }
                     Spacer()
                 }
             }
-        }.backgroundPreferenceValue(FirstNonNilPreferenceKey<Anchor<CGRect>>.self) { boundsAnchor in
+        }
+        .backgroundPreferenceValue(FirstNonNilPreferenceKey<Anchor<CGRect>>.self) { boundsAnchor in
```
skywalkerdude, line 75: could you remove this line break

skywalkerdude:
I think you need to update the snapshot tests btw

skywalkerdude:
Looks like the snapshot test found a regression:

https://github.com/HymnalDreamTeam/Hymns-iOS/blob/34c363dffa882382c856d5da945ba48e2ceb85d2/HymnsSnapshotTests/__Snapshots__/BrowseCategoriesSnapshots/test_browseView.1.png

I didn't realize this earlier, but it looks like this change drastically breaks the browse tab bar :(

finben:
I thought we already tried to go the route of using geometry reader but always had issues with it actually lining up the scroll view. Remember that? We could never properly center the scroll view with geo reader I thought?

skywalkerdude:
Yeah, that's when we only set the width to the width of the geometry proxy. But I'm thinking maybe we can try to check if the width exceeds the geometry proxy or not? I dunno. I'll give it a stab and see if it's viable.

skywalkerdude:
@finben we can close this now right?

PR #365, opened 2020-07-01
Fix the tab bar spacing issue by dynamically calculating the width before setting the frame's width

skywalkerdude:
A rather hacky solution, but seems to be the best solution until Apple fixes their scroll view .

Please read this and let me know if this is clear:

> First we calculate the width of all the tabs by putting them into a ZStack. We do this in order to set the width of the eventual HStack appropriately. If the total width is less than then width of the containing GeometryProxy, then we should set the width to the width of the geometry proxy so that the tabs take up the entire width and are equaly spaced. However, if the total width is greater than or equal to the width of the containing GeometryProxy, then we should set the frame's width to nil to allow it to scroll offscreen.


skywalkerdude:
@finben while reviewing this, could you also check this on different screen sizes and font sizes to make sure it looks right? I tried it on some of mine and it seems to look fine for me

skywalkerdude:
wow, good find! Yeah you are right. I don't think it's because of performance, since the page never really loads. There was a misconfiguration where it was throwing away the old view model and using a new one, so the data never appeared. I fixed it with my most recent commit!

finben:
@skywalkerdude Okay, this is alot better. Still when in landscape if you tab between options you get a wierd behavior where the view is like animating off to the side as the new one comes in. Try going to landscape mode then browse and shifting between tabs. Let me know if you need me to send a video.

skywalkerdude:
another good find. I think I fixed it in the recent commit!

skywalkerdude:
closes https://github.com/HymnalDreamTeam/Hymns-iOS/issues/322

Issue #320, opened 2020-06-17
Tags too close to the tabs in Browse


Emyeh:
Needs to have 16dp padding between bottom of tabs and the tag.

![IMG_7579](https://user-images.githubusercontent.com/26554065/84859386-e758fe80-b021-11ea-980a-ed0f8ead7671.PNG)

PR #372, opened 2020-07-05
Add padding to the top the tab list and change radius for tab pills

skywalkerdude:
fixes https://github.com/HymnalDreamTeam/Hymns-iOS/issues/320


Emyeh:
test_manyTags.1.png has too much padding between the bottom of the tags and the CTAs

PR #374, opened 2020-07-05
Try to use the NaivgationStackView to manage the back stack.

skywalkerdude:
Let's see if this works... (https://github.com/biobeats/swiftui-navigation-stack/blob/master/Sources/NavigationStack/NavigationStack.swift)

@finben could you try to play around with this branch a bit? The view transitions don't work too well for me, but maybe you can edit the transition animations and have better luck than I did (https://github.com/biobeats/swiftui-navigation-stack#usage)


finben:
Also, note that this branch reintroduces the tab bar icon alignment
Simulator Screen Shot - iPhone 11 Pro - 2020-07-06 at 11 10 01
 bug.

skywalkerdude:
One caveat: currently, NavigationStackView calls onAppear immediately after push/pop, versus after the transition is completed. (https://github.com/biobeats/swiftui-navigation-stack/issues/19). If you add a 0.2 second delay in the onAppear before doing the viewModel logic, the animation transition is perfect: https://github.com/biobeats/swiftui-navigation-stack/issues/23#issuecomment-652881759

@finben what do you think? should we go with this workaround? Or should we just wait until the library is fixed or SwiftUI decides to support navigation stack management?

skywalkerdude:
We do need to fix the tab bar icon alignment bug. @finben could you take a stab to see if you can fix it on this branch?

PR #381, opened 2020-07-10
There is a wierd flash but other than that it looks like a solution

finben:
Besides the wierd flash when you route back, I think it looks pretty good. This is actually the behavior of your android app I think.


skywalkerdude:
What weird flash are you referring to?

PR #382, opened 2020-07-13
Format some things for landscape

skywalkerdude:
1) Force stack navigation view for initial release. Can iterate and make ipad configurations in the future
2) Move verses to be center-aligned, as per https://github.com/HymnalDreamTeam/Hymns-iOS/issues/135

(fixes https://github.com/HymnalDreamTeam/Hymns-iOS/issues/135)


skywalkerdude:
@Emyeh let me know if the screenshot looks okay!

skywalkerdude:
I don't think it's off-centered? If you look at where "redemption" ends on the 4th verse and the start of the verse numbers, at least to me, they look like the same padding. I think maybe it just looks off-centered because the lines have different lengths and if you want to keep the verses themselves left-aligned, some lines are going to look off-centered. Does that make sense? Do you think that's an ok state to be in? Or should we try to fix it somehow?

As for the buttons, there are only 3 buttons cuz I set it to only have 3 buttons for that particular screenshot. I'll fix the off-centered-ness in another pull request

Emyeh:
No it's fine as it is.

It would be good to keep the bottom buttons centered regardless of the # of buttons there. That way it's aligned with the centered hymn stanzas.

Issue #378, opened 2020-07-08
Back caret hard to tap


Emyeh:
![IMG_7710](https://user-images.githubusercontent.com/26554065/86980284-4fd65f80-c138-11ea-8b79-3a3e71b598d8.PNG)

I've run into this issue a lot. My finger is small but it takes me several taps to go back. Not sure if the tap target isn't 48x48dp or it's a bug with the caret itself.

skywalkerdude:
Yeah, I've run into this also. It is certainly quite the annoyance. @finben could you see if you can make the touch target larger?

PR #383, opened 2020-07-15
Padding within the rectangle adds to tappable space

finben:
Here is a before and after screenshot with the tappable space. Addresses #378 

PS. For some reason my images are in the wrong order. Obviously, the bigger green space is the after my edits photo lol

Simulator Screen Shot - iPhone 11 Pro - 2020-07-15 at 15 12 16

Simulator Screen Shot - iPhone 11 Pro - 2020-07-15 at 15 08 38


skywalkerdude:
It seems like in https://github.com/HymnalDreamTeam/Hymns-iOS/pull/232, the padding should have been put on the label instead of on the button. That way the entire padding will be part of the touch target versus just the small label part.

Issue #321, opened 2020-06-17
Hymns titles in under Scripture in Browse aren't aligned


Emyeh:
I see that the verse number is aligned but the hymn title alignment starts to look funny. Can we give a max length for the numbers and align the hymn titles with 16dp padding between?

IMG_7581

skywalkerdude:
@Emyeh I don't see the screenshot you posted?

skywalkerdude:
oh, I see. This happens because the verse reference takes a different amount of space. How do you think we should align it? What if the verse reference is too long?

Emyeh:
We can try something like this 
![image](https://user-images.githubusercontent.com/26554065/86565363-8a63b080-bf1c-11ea-99fd-0aa977c8a26b.png)

- Keep the verse # the same text size and left align it with the book title
- Add 16dp padding between the verse number and hymn title
- Remove the left caret for more space on the left

PR #384, opened 2020-07-18
Align hymn titles under Scripture in browse

skywalkerdude:
Fixes https://github.com/HymnalDreamTeam/Hymns-iOS/issues/321


skywalkerdude:
@Emyeh let me know if this looks okay to you. The last two are "accessibility sizes" which basically means they're extra big. In that case, I created an alternate layout where they are vertically stacked instead of horizontally stacked.

Emyeh:
This is much better! I think this is good enough to be a11y compliant, all the tap targets look accessible now.

Issue #262, opened 2020-06-09
Add toast to let the user know their feedback email was sent. 


finben:
Needed in Feedback under settings. 
Would it be possible to show like a simple toast message telling them that the email succeeded or failed?

PR #385, opened 2020-07-19
Add a toast message for when feedback emails are sent, canceled, saved, or failed

skywalkerdude:
fixes https://github.com/HymnalDreamTeam/Hymns-iOS/issues/262


skywalkerdude:
Light mode:
Screen Shot 2020-07-19 at 11 03 19 AM

Dark mode:
Screen Shot 2020-07-19 at 11 03 57 AM

@Emyeh do these toast messages look okay to you? Let me know if you want them to look different/design new app-wide toasts. I think we have a lot of control over how these toasts look

Emyeh:
Let's do something like this 
image

The other states are in the figma file

PR #390, opened 2020-07-23
Translation into German

finben:
I sent a copy to Jonathen Gil to proofread by translations. But here is what I came up with so far. One todo is that all the titles of the Bible books need to be added to our localized strings. Everything else is looking good though in my simulator for German.


Hymns/Browse/BrowseTab.swift
```diff
@@ -16,19 +16,19 @@ extension BrowseTab {
     var label: String {
         switch self {
         case .tags:
-            return "Tags"
+        return NSLocalizedString("Tags", comment: "")
```
skywalkerdude, line 19: What is the "comment" used for?

skywalkerdude: Ah, I see that it'll get auto-ported to the Localizable file. Can we go through and add them in to make it easy to translate? 😛 I'll help with it too when I get time  https://developer.apple.com/documentation/foundation/nslocalizedstring

finben: Do you mean add the same string to the comment? If that is what you mean I did that below.

skywalkerdude:
@finben I fixed it up a little bit and added some new strings that we added since you last committed. Could you translate those also, then submit?

PR #392, opened 2020-07-24
Fix the infinite-loading audio player issue by switching to use AVAudioPlayer instead of AVPlayer

skywalkerdude:
Fix the infinite-loading audio player issue by switching to use AVAudioPlayer instead of AVPlayer and managing the downloading/buffering myself instead of relying on AVPlayer to do it. The upside here is that we now have access to the duration, and can introduce the scrubber back in!

@finben let me know if you want to take a stab at bringing the slider back. I believe we have the duration now in `player.duration`


finben:
Yes, I'll take a stab! Awesome! I think last time I tried this the duration it always gave me was 0. But yeah I'd love to check it out. Good work!

PR #399, opened 2020-07-25
Sound cloud option2

finben:
Let me know what you think of the functionality here versus my other branch. The cool thing about taking this route of using a ZStack and hiding is that the user can play something on SoundCloud then go back to the lyrics or chords.

The one downside is sheets looks a little bit smoothing imo 

To me the gain in functionality of being able to play a song then go back to lyrics and chords is worth it.


Hymns.xcodeproj/project.pbxproj
```diff
@@ -2082,7 +2090,7 @@
 				CODE_SIGN_STYLE = Automatic;
 				CURRENT_PROJECT_VERSION = 12;
 				DEVELOPMENT_ASSET_PATHS = "\"Hymns/Preview Content\"";
-				DEVELOPMENT_TEAM = VBX3S26MR4;
+				DEVELOPMENT_TEAM = N2YVF8J6Z5;
```
skywalkerdude, line 2093: pls revert all of these

finben:
No longer using this PR

PR #406, opened 2020-07-26
Youtube, deeplinking

finben:
Obviously, we can change the ui side. I just wanted to see how the functionality worked here. Lmk what you think of the deeplink approach.


finben:
Migrated to #410

PR #410, opened 2020-07-26
Moving SoundCloud from deeplink back to webView

finben:
This will replace my other two PR's.

As for the upsell banner. I am just leaving it out. The thing is that it doesn't appear to be working even when you click into a song. It never works so might as well leave it out. The nice thing is the Soundcloud logo is there though when you click into a song.


Hymns/Common/SoundCloudWebView.swift
```diff
@@ -0,0 +1,66 @@
...
+            parent.loadStatusChanged?(true, nil)
+        }
+
+        func webView(_ webView: WKWebView, didFinish navigation: WKNavigation!) {
+
+            parent.isLoading = false  //Used for activity indicator
+            webView.evaluateJavaScript("document.querySelector('.app__upsell').remove();", completionHandler: { (_, _) -> Void in
+            })
+//            webView.evaluateJavaScript("document.querySelector('.app__header').remove();", completionHandler: { (_, _) -> Void in
+//            })
```
skywalkerdude, line 46: comment?

Hymns/Infrastructure/Data/Models/UiHymn.swift
```diff
@@ -30,6 +30,12 @@ struct UiHymn: Equatable {
         return title + self.title
     }
 
+    var searchableTitle: String {
+        var title = self.title
+        title = title.replacingOccurrences(of: " ", with: "+", options: .literal, range: nil)
+        return title
+    }
+
```
skywalkerdude, line 38: I don't think you need this. Just use `toEncodedUrl` to encode it correctly

Hymns/Common/SoundCloudWebView.swift
finben: I wasn't sure if we would ever want it later lol... It was a pain to get that javascript name .app__header. But I guess we can delete it?

Hymns/Common/SoundCloudWebView.swift
```diff
@@ -0,0 +1,66 @@
+import SwiftUI
+import WebKit
+
+struct SoundCloudWebView: UIViewRepresentable {
```
skywalkerdude, line 4: This doesn't build for me because SoundCloudWebView is not added in the project.pbxproj

Hymns/Display/DisplayHymnBottomBar.swift
```diff
@@ -4,6 +4,8 @@ import SwiftUI
 struct DisplayHymnBottomBar: View {
 
     @Binding var dialogBuilder: (() -> AnyView)?
+    @Binding var toggleSoundCloud: Bool
+    @Binding var initiatedSoundCloud: Bool
```
skywalkerdude, line 8: What is the benefit of passing all this stuff in and displaying it over the hymn lyrics versus just making it a sheet that comes up? I think that would simplify the code a lot.

just do 

```
        case .soundCloud(let url):
            self.sheet = .soundCloud(url)
```

Hymns/Common/SoundCloudWebView.swift
finben: That's wierd....

Hymns/Display/DisplayHymnBottomBar.swift
finben: The problem is then we have no way to play soundCloud in the background. As soon as you pull the sheet back down your sound cloud session is over. The benefit of this way, even though admittedly it is not as pretty in the code, is that it makes it easy for someone to seamlessly shift between enjoying listening to the song alongside either singing the lyrics or playing along with guitar.

Hymns/Common/SoundCloudWebView.swift
```diff
@@ -0,0 +1,66 @@
...
+            })
+
+            webView.evaluateJavaScript("document.querySelector('.footer.show').remove();", completionHandler: { (_, _) -> Void in
+            })
+
+            webView.evaluateJavaScript("document.querySelector('.footer').remove();", completionHandler: { (_, _) -> Void in
+            })
+
+            webView.evaluateJavaScript("document.querySelector('.categories').remove();", completionHandler: { (_, _) -> Void in
+            })
```
skywalkerdude, line 57: What do all these others things do? It looks like the .app__upsell is the only thing that we want to remove right? Or is there other stuff too?

finben: The other stuff is nice to remove. Without removing it we only get to see like the top 5 songs searched and then we get a big footer banner with another upsell thing saying to see more songs you need to use the app

Hymns/Infrastructure/Data/Models/UiHymn.swift
finben: Nice, did this in my last commit. Thanks

Hymns/Common/SoundCloudWebView.swift
finben: Should be fixed now. I recommitted, that was super wierd.

finben:
I can't figure out why I am failing the tests? It is passing on my local...

finben:
Note, this needs to be tested on actual device because it relys on a call to AV.sharedInstance.isotheraudioplaying which will always be false on the simulator!

Hymns/Common/SoundCloudWebView.swift
```diff
@@ -0,0 +1,65 @@
...
+
+        func webView(_ webView: WKWebView, didCommit navigation: WKNavigation!) {
+            parent.loadStatusChanged?(true, nil)
+        }
+
+        func webView(_ webView: WKWebView, didFinish navigation: WKNavigation!) {
+
+            parent.isLoading = false  //Used for activity indicator
+
+            webView.evaluateJavaScript("document.querySelector('.upsellBanner__appButton').innerHTML = 'SoundCloud Results';", completionHandler: { (_, _) -> Void in
```
skywalkerdude, line 44: If you're going to change the top bar, can you remove the click listener too? Now it still looks like clicking it brings the popup up

Hymns/Display/DisplayHymnView.swift
```diff
@@ -27,9 +35,29 @@ struct DisplayHymnView: View {
                     } else {
                         viewModel.currentTab.content
                     }
+                    if !self.showSoundCloud && self.initiatedSoundCloud && AVAudioSession.sharedInstance().isOtherAudioPlaying {
```
skywalkerdude, line 38: Is there a way to make it listen to when playback ends?

Hymns/AppDelegate.swift
```diff
@@ -1,4 +1,4 @@
-import Firebase
+import FirebaseCore
```
skywalkerdude, line 1: Is this needed?

Hymns.xcodeproj/project.pbxproj
```diff
@@ -483,18 +484,19 @@
...
+		897D3DC4446064BA5F96D6F5 /* Pods-Hymns.release.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-Hymns.release.xcconfig"; path = "Target Support Files/Pods-Hymns/Pods-Hymns.release.xcconfig"; sourceTree = "<group>"; };
+		900AC72914643D3EA3DCABC2 /* Pods-HymnsTests.release.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-HymnsTests.release.xcconfig"; path = "Target Support Files/Pods-HymnsTests/Pods-HymnsTests.release.xcconfig"; sourceTree = "<group>"; };
 		B51BA5C82430075300CA52F7 /* DisplayHymnView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = DisplayHymnView.swift; sourceTree = "<group>"; };
+		B55B18B624CE38AE0087BB7D /* SoundCloudWebView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SoundCloudWebView.swift; sourceTree = "<group>"; };
+		B55B18B824CE38E60087BB7D /* SoundCloudView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SoundCloudView.swift; sourceTree = "<group>"; };
```
skywalkerdude, line 499: What did you do to your project settings? Why are there so many changes in this file?

Usually when I add 2 files, this is all there is and not all these other changes

Hymns/Common/SoundCloudWebView.swift
```diff
@@ -0,0 +1,65 @@
...
+            })
+
+            webView.evaluateJavaScript("document.querySelector('.footer.show').remove();", completionHandler: { (_, _) -> Void in
+            })
+
+            webView.evaluateJavaScript("document.querySelector('.footer').remove();", completionHandler: { (_, _) -> Void in
+            })
+
+            webView.evaluateJavaScript("document.querySelector('.categories').remove();", completionHandler: { (_, _) -> Void in
+            })
```
skywalkerdude, line 56: can you at least add comments saying what each of these does? That way we're doing things intentionally and not just adding code without knowing why

Hymns/Display/DisplayHymnView.swift
```diff
@@ -16,6 +19,11 @@ struct DisplayHymnView: View {
             if !viewModel.isLoaded {
                 ActivityIndicator().maxSize()
             } else {
+                ZStack {
+                    if self.initiatedSoundCloud {
```
skywalkerdude, line 23: Can you have consistent naming of variables? Pick either soundCloudinitiated or initiatedSoundCloud (although frankly both of them are kind of bad lol. "initiated" isn't really an accurate description of what it's doing. Maybe `isSoundCloudBackgrounded` is more informative

Hymns/Display/DisplayHymnBottomBar.swift
```diff
@@ -4,6 +4,8 @@ import SwiftUI
 struct DisplayHymnBottomBar: View {
 
     @Binding var dialogBuilder: (() -> AnyView)?
+    @Binding var toggleSoundCloud: Bool
```
skywalkerdude, line 7: I think we don't need this variable. We already have dialogBuilder to pass information from the bottom bar to the container, so can you just use that instead

Hymns/Display/DisplayHymnViewModel.swift
```diff
@@ -10,6 +10,7 @@ class DisplayHymnViewModel: ObservableObject {
 
     @Published var isLoaded = false
     @Published var title: String = ""
+    @Published var searchTitle: String = ""
```
skywalkerdude, line 13: Can you just use `computedTitle`

Hymns/Display/SoundCloudView.swift
```diff
@@ -0,0 +1,44 @@
...
+                    Button(action: {
+                        self.showSoundCloud.toggle()
+                    }, label: {
+                        Image(systemName: "minus.square.fill").accentColor(.accentColor)
+                    })
+                    Spacer()
+                }.padding()
+                ZStack {
+                    SoundCloudWebView(isLoading: self.$isLoading, url: url).eraseToAnyView().opacity(isLoading ? 0 : 1)
+                    ActivityIndicator().maxSize().eraseToAnyView().opacity(isLoading ? 1 : 0)
```
skywalkerdude, line 32: What if you didn't have this? I see like 2-3 activity indicators when trying to load the webview, which is a bit weird. It's not a blocker, but can we reduce this to only show 1 indicator?

Hymns/Display/SoundCloudView.swift
```diff
@@ -0,0 +1,44 @@
+import SwiftUI
+
+struct SoundCloudView: View {
+    @State private var isLoading = true
+    @Binding var showSoundCloud: Bool
+    @Binding var soundCloudinitiated: Bool
```
skywalkerdude, line 6: If we are just looking at `isOtherAudioPlaying` then we don't need this right?

Hymns/Display/DisplayHymnView.swift
skywalkerdude, line 38: from the docs:

```    /* Will be true when another application is playing audio.
    Note: As of iOS 8.0, Apple recommends that most applications use secondaryAudioShouldBeSilencedHint instead of this property.
    The otherAudioPlaying property will be true if any other audio (including audio from an app using AVAudioSessionCategoryAmbient)
    is playing, whereas the secondaryAudioShouldBeSilencedHint property is more restrictive in its consideration of whether 
    primary audio from another application is playing.  
    For additional information, see https://developer.apple.com/library/content/qa/qa1882/_index.html
    */
    @available(iOS 6.0, *)
    open var isOtherAudioPlaying: Bool { get }
```
pls fix.

Hymns/Common/SoundCloudWebView.swift
finben: I can try. I guess the question is should we even change it anymore. I think the reason I was looking to change it was that we didn't think the button worked. But after testing on a real device, it does seem like the button works and lets you got to the app. What do you think? Better to change and disable or leave in?

Hymns/AppDelegate.swift
finben: So for whatever reason my XCode keeps throwing me an error about this when I use Firebase now. It gives me a red error flag but it actually does not stop the build. This started happening to me awhile back. I don't know if it was because our pods were updated or what. I can revert though. You don't get any error flags?

Hymns.xcodeproj/project.pbxproj
finben: It's because I had to do alot with my pods. I didn't commit all the pod stuff I was doing updating, installing etc. This shouldn't matter though right? Other than just being annoying

Hymns/Display/DisplayHymnView.swift
finben: Yeah, I had seen this. The problem is that with a WebView it does not change the bool secondaryAudioShouldBeSilencedHint. So that bool will always be false whether we have music playing or not.   

However, we may want to include some logic with this. To distinguish when outside music is playing and contrast that with isOtherAudioPlaying to discern whether it is our WebView playing the other audio or an outside app

Hymns/Display/SoundCloudView.swift
finben: I think we still do. The reason why is we need a way to actually close the webView and stop the music. So let's say that the user open SoundCloud, plays music, goes back to the lyrics. Music is playing and then they want to stop that music from playing. So they click our X on the banner. We need to be able to close out that webView.   

That is what this var does

Hymns/Display/SoundCloudView.swift
finben: I reduced this one here. That brings us from 3 to 2. I'm not sure if I can reduce another. I'll try

finben: I think it's because one is coming from the webKit and the other is from the soundcloud site itself

Hymns/Display/DisplayHymnViewModel.swift
finben: ComputedTitle is like Hymn 608. So we wouldn't want to use that for what Sound cloud searches or we will never find what we are looking for

Hymns/Display/DisplayHymnBottomBar.swift
finben: The problem is I use this variable when the user taps on the soundCloud button. If I use dialogBuilder I am not sure how that would work. So if the user clicks the soundCloud button I pass a view up?

Hymns/Display/DisplayHymnView.swift
finben: Thanks! great feedback. I went ahead and made the change here in recent commit

Hymns/Common/SoundCloudWebView.swift
finben: Done in recent commit

Hymns/Display/DisplayHymnView.swift
finben: I know... We need this. I wish it would flip our isOtherAudioPlaying variable between songs. Unfortunately, it just keeps playing right into the next song. I'm not sure if there is something fired on the website that we can listen for using JavaScript. That is what we would have to do I think.

Hymns/Assets.xcassets/SoundCloudIcon.imageset/Contents.json
```diff
@@ -0,0 +1,21 @@
...
+      "idiom" : "universal",
+      "scale" : "1x"
+    },
+    {
+      "idiom" : "universal",
+      "scale" : "2x"
+    },
+    {
+      "idiom" : "universal",
+      "scale" : "3x"
```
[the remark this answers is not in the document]
finben: Waiting on Emily to generate for us

Hymns/Common/SoundCloudWebView.swift
skywalkerdude: If it works well, we should leave it. I don't think it works well though. The "Download Soundcloud" button doesn't work for me. Does it work for you?

Hymns/AppDelegate.swift
skywalkerdude: yeah it doesn't cause any errors for me lol

Hymns.xcodeproj/project.pbxproj
skywalkerdude: Oh, right it was the pod update. I forgot about that. okay, that's fine then. I just didn't know what it was.

Hymns/Display/DisplayHymnViewModel.swift
skywalkerdude: I think computed title is Hymn 608: TITLE_OF_HYMN. Can you try it out to see how good soundcloud's search algorithm is? I thought it'd be even better because it searches both the title of the hymn and Hymn 608 so we get both results. But that might not be true

Hymns/Common/SoundCloudWebView.swift
skywalkerdude: If it works well, we should leave it. But I don't think it works well though. The "Download Soundcloud" button doesn't work for me. Does it work for you?

Hymns/Display/DisplayHymnView.swift
skywalkerdude: Okay, that's fine then

skywalkerdude: > However, we may want to include some logic with this. To distinguish when outside music is playing and contrast that with isOtherAudioPlaying to discern whether it is our WebView playing the other audio or an outside app

How do you think we should distinguish this?

Hymns/Display/SoundCloudView.swift
skywalkerdude: Oh, is the `webKit` one also not controlled by us?

Hymns/Display/DisplayHymnBottomBar.swift
skywalkerdude: Okay, let me play around with this and see if I can get it. I thought of a potential solution where we created a DialogViewModel with customizable properties and pass that in to the Dialog. That way, we have more control over the dialog and we can also introduce a callback for what to do when the dialog is dismissed. I'll try to find time to play with this tomorrow.

PR #412, opened 2020-07-26
Change overflow icons to 6 in number

finben:
Some controversial changes. We can discuss. 

Change overflow to 6 as we talked about yesterday. make song info last, and change icon to match others for youtube.


Hymns/Display/DisplayHymnBottomBarButton.swift
```diff
@@ -80,7 +80,7 @@ extension BottomBarButton {
         case .soundCloud:
             return BottomBarLabel(imageName: "cloud", a11yLabel: label).foregroundColor(.primary)
         case .youTube:
-            return BottomBarLabel(imageName: "play.rectangle.fill", a11yLabel: label).foregroundColor(.primary)
+            return BottomBarLabel(imageName: "play.rectangle", a11yLabel: label).foregroundColor(.primary)
```
skywalkerdude, line 83: I think that play.rectangle.fill is closer to Youtube's icon, which is a filled rectangle

finben: I know, lol I figured that's what you were thinking behind the choice. To me it just looks off ascetically, since it is the only one filled. I feel like the viewers attention immediately gets drawn to it over the other icons. Do you want to run it by Emily and see what she thinks? I guess she probably wants us to use the figma specs instead of SF symbols anyways, although I'm not really sure how to convert them.

skywalkerdude: I can see your point. I think you're right. Maybe the non-filled one is better than. @Emyeh could you take a look at the screenshots and weigh in?

PR #413, opened 2020-07-26
Let's get Lottie in for launch. No reason not to

finben:
Revisit #274 which was basically already ready to be pushed in but just stagnated. It would be really nice to have this animation working when people start downloading our app for the best experience.


Hymns/Home/LaunchRouterView.swift
```diff
@@ -0,0 +1,26 @@
...
+                }.eraseToAnyView()
+            } else {
+                return HomeContainerView().eraseToAnyView()
+            }
+        }.onAppear {
+            if !UserDefaults.standard.bool(forKey: "didLaunchBefore") {
+                UserDefaults.standard.set(true, forKey: "didLaunchBefore")
+                self.showSplash = true
+            }
+        }
```
skywalkerdude, line 24: Can you use the `UserDefaultsManager` to do this? I created a property wrapper so you can probably just use that instead of doing this in onAppear: https://swiftsenpai.com/swift/create-the-perfect-userdefaults-wrapper-using-property-wrapper/

Hymns/Home/LaunchRouterView.swift
```diff
@@ -0,0 +1,26 @@
+import SwiftUI
+
+struct LaunchRouterView: View {
+    @State var showSplash: Bool = false
+
+    var body: some View {
+        Group { () -> AnyView in
+            if showSplash {
+                return SplashScreenView()
+                    .opacity(showSplash ? 1 : 0)
```
skywalkerdude, line 10: Do we need this opacity? If showSplash is false, we won't even show this right

Hymns/SceneDelegate.swift
```diff
@@ -19,7 +19,13 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
 
         // Create the SwiftUI view and set the context as the value for the managedObjectContext environment keyPath.
         // Add `@Environment(\.managedObjectContext)` in the views that will need the context.
+        #if !DEBUG
+        let rootView = LaunchRouterView().environment(\.managedObjectContext, context)
+        #endif
+
+        #if DEBUG
         let rootView = HomeContainerView().environment(\.managedObjectContext, context)
+        #endif
```
skywalkerdude, line 28: What's this for?

Hymns/Home/LaunchRouterView.swift
```diff
@@ -0,0 +1,26 @@
...
+                    .onAppear {
+                        DispatchQueue.main.asyncAfter(deadline: .now() + 2.7) {
+                            self.showSplash = false
+                        }
```
skywalkerdude, line 14: Is it possible to use `SplashScreenView`'s coordinator to determine when the animation is finished instead of waiting 2.7 seconds? I don't think there's any guarantee that the animation will run in 2.7 seconds, is there?

Hymns/Display/SplashScreenView.swift
```diff
@@ -0,0 +1,28 @@
+import SwiftUI
+import Lottie
+
+//Implemented using https://www.raywenderlich.com/4503153-how-to-create-a-splash-screen-with-swiftui
+//And the help from https://www.youtube.com/watch?v=fVehE3Jf7K0
+struct SplashScreenView: UIViewRepresentable {
+    let animationView = AnimationView()
+    var filename = "firstLaunchAnimation"
```
skywalkerdude, line 8: Is this ever edited? If not, can this be a let

Hymns/Display/SplashScreenView.swift
```diff
@@ -0,0 +1,28 @@
...
+
+        NSLayoutConstraint.activate([
+            animationView.heightAnchor.constraint(equalTo: view.heightAnchor),
+        animationView.widthAnchor.constraint(equalTo: view.widthAnchor)])
+
+        return view
+    }
+
+    func updateUIView(_ uiView: UIViewType, context: UIViewRepresentableContext<SplashScreenView>) {
+        }
```
skywalkerdude, line 27: Indent

Hymns/Display/SplashScreenView.swift
```diff
@@ -0,0 +1,28 @@
+import SwiftUI
+import Lottie
+
+//Implemented using https://www.raywenderlich.com/4503153-how-to-create-a-splash-screen-with-swiftui
+//And the help from https://www.youtube.com/watch?v=fVehE3Jf7K0
+struct SplashScreenView: UIViewRepresentable {
+    let animationView = AnimationView()
```
skywalkerdude, line 7: Just curious, why is this an instance var instead of just a local variable? Does makeUIView get called multiple times and we need the same instance of animation view?

Hymns/SceneDelegate.swift
finben: This keeps the lottie animation from ever running in debug mode at all. That is what kept breaking all of our tests like anything that was time based. This keeps it from breaking things like ui tests, probably snapshot tests too. So we only run the Lottie animation when not in debug mode

Hymns/Home/LaunchRouterView.swift
finben: Nice, good catch.

Hymns/Home/LaunchRouterView.swift
finben: Yes, there is. 2.7 seconds is the amount of time that it always takes to run. It should never change. and is set. Every-time, I have tested it finishes right on time.

Hymns/SceneDelegate.swift
skywalkerdude: Ahh, I didn't see the "!" in fromt of the DEBUG. That's cool!

Hymns/Home/LaunchRouterView.swift
skywalkerdude: But I feel like when we test, it's in a rather controlled environment, with the same amount of ram, with all other apps cleared, and on a simulator/same device over and over again. Whereas in the wild, things will be quicker and slower depending on a number of factors (https://www.knowyourmobile.com/user-guides/4-ways-to-speed-up-slow-iphone/)

I guess if it's not possible to create a listener with a Coordinator, then I'd be okay with keeping the 2.7 seconds, since this isn't a huge deal anyway. But just want to make sure that we have to do it this way before hardcoding a magic number (https://en.wikipedia.org/wiki/Magic_number_%28programming%29) in here

skywalkerdude:
I'm not sure what's going on with the build. It looks like it's not even building? But it looks like it's synced with master right

skywalkerdude:
oh, it looks like you deleted the mockingbird files
Screen Shot 2020-07-26 at 7 59 59 PM

finben:
wth... I have been having so much wierd stuff happening today. My pbxproj not containing files, and now somehow deleting mocking files?

finben:
Ohh that probably happened when I updated pods or something to add the pod for lottie

Hymns/Home/LaunchRouterView.swift
finben: Yes, I did this. Can you just check that I did it right. The changes are on my new Lottie PR

Hymns/Display/SplashScreenView.swift
finben: Changed in new PR

Hymns/Display/SplashScreenView.swift
finben: Done in new PR

Hymns/Display/SplashScreenView.swift
finben: Changed to local var. Good point.

PR #414, opened 2020-07-27
Change overlow limit to 6 and reorder bottom bar icons

finben:
Just did another PR. I think it was just that I hadn't updated the unit tests until now though. Crossing my fingers that this passes tests!


HymnsTests/Display/DisplayHymnBottomBarViewModelSpec.swift
```diff
@@ -175,10 +172,10 @@ class DisplayHymnBottomBarViewModelSpec: QuickSpec {
...
                     expect(target.buttons[1]).to(equal(.fontSize))
                     expect(target.buttons[2]).to(equal(.tags))
-                    expect(target.buttons[3]).to(equal(.songInfo(SongInfoDialogViewModel(hymnToDisplay: classic1151))))
-                    expect(target.buttons[4]).to(equal(.soundCloud(URL(string: "https://soundcloud.com/search?q=title")!)))
-                    expect(target.buttons[5]).to(equal(.youTube(URL(string: "https://www.youtube.com/results?search_query=title")!)))
+                    expect(target.buttons[3]).to(equal(.soundCloud(URL(string: "https://soundcloud.com/search?q=title")!)))
+                    expect(target.buttons[4]).to(equal(.youTube(URL(string: "https://www.youtube.com/results?search_query=title")!)))
                     expect(target.overflowButtons).to(beNil())
+                    expect(target.buttons[5]).to(equal(.songInfo(SongInfoDialogViewModel(hymnToDisplay: classic1151))))
```
skywalkerdude, line 178: nit: move this 1 line up so it's group withed the other calls

HymnsTests/Display/DisplayHymnBottomBarViewModelSpec.swift
```diff
@@ -140,15 +140,12 @@ class DisplayHymnBottomBarViewModelSpec: QuickSpec {
                 }
                 let mp3Url = URL(string: "http://www.hymnal.net/en/hymn/h/1151/f=mp3")!
                 it("should have \(DisplayHymnBottomBarViewModel.overflowThreshold) buttons in the buttons list and nothing in the overflow") {
-                    expect(target.buttons).to(haveCount(7))
+                    expect(target.buttons).to(haveCount(5))
```
skywalkerdude, line 143: Can you please update this test so that it has 6 elements? Then assert that the overflow is nil. It's meant to test the case where you hit the max number of options without going to overflow.

finben: Lol, I can't seem to get this right. So annoying, no matter what I add or subtract I am getting 5 elements. I don't see how before I changed it we had 7 and now we have 5.

HymnsTests/Display/DisplayHymnBottomBarViewModelSpec.swift
finben: resolved

HymnsTests/Display/DisplayHymnBottomBarViewModelSpec.swift
skywalkerdude: ok, I fixed it. It looks like the minimum number of buttons we have now is 6, so we can just remove this test and use the below test instead.

PR #417, opened 2020-07-27
Add initial Lottie launch animation (only shows on first install open)

finben:
Migrating from #413 ... Please see #413 for my comments on your PR comments. 

Unfortunately, I had to move PR's and start another one because it was such a headache to get that mockingbird file back in. I just started over and did that here.


Hymns/SceneDelegate.swift
```diff
@@ -19,7 +19,13 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
 
         // Create the SwiftUI view and set the context as the value for the managedObjectContext environment keyPath.
         // Add `@Environment(\.managedObjectContext)` in the views that will need the context.
+        #if !DEBUG
+        let rootView = LaunchRouterView().environment(\.managedObjectContext, context)
+        #endif
+
+        #if DEBUG
         let rootView = HomeContainerView().environment(\.managedObjectContext, context)
+        #endif
```
skywalkerdude, line 28: 1) You can do use #else for the else case instead of if DEBUG and if !DEBUG
2) This thing makes me nervous. Since we usually develop on the debug build, we won't ever catch any errors that appear for the LaunchRouterView and the animation stuff until it launches to production. Could you find another way to determine if you're in a test or not? You can try to inspect `CommandLine.arguments` or something to see if there's any way we can tell between a debug test build and a debug dev build.

skywalkerdude: Actually, it looks like you can add your own preprocessor macro to the build settings: https://stackoverflow.com/questions/242817/xcode-preprocessor-macros

So you can eventually just do
#if TEST
...
#else
...
#endif

finben: Okay, according to this stackoverflow preprocessor won't work  for this. https://stackoverflow.com/questions/6748087/xcode-test-vs-debug-preprocessor-macros 

but I did come up with a solution, that I think will work. It checks if testing instead of Debug and then routes accordingly.

Hymns/SceneDelegate.swift
```diff
@@ -19,12 +19,17 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
 
         // Create the SwiftUI view and set the context as the value for the managedObjectContext environment keyPath.
         // Add `@Environment(\.managedObjectContext)` in the views that will need the context.
-        let rootView = HomeContainerView().environment(\.managedObjectContext, context)
+        let rootView = LaunchRouterView().environment(\.managedObjectContext, context).eraseToAnyView()
+        let rootViewTest = HomeContainerView().environment(\.managedObjectContext, context).eraseToAnyView()
+
+        var testing: Bool {
+            return ProcessInfo.processInfo.environment["XCTestConfigurationFilePath"] != nil
+        }
```
skywalkerdude, line 27: Why is this a computed variable? why not just

`var isTesting = ProcessInfo.processInfo.environment["XCTestConfigurationFilePath"] != nil`

then just do
`let rootView = isTesting ? HomeContainerView().environment(\.managedObjectContext, context).eraseToAnyView() : LaunchRouterView().environment(\.managedObjectContext, context).eraseToAnyView()`

skywalkerdude: It looks like this case doesn't cover UI tests

skywalkerdude: Also, I would surround this whole thing with #IF DEBUG just in case

Issue #423, opened 2020-07-28
Hide certain options if network is unavailable


skywalkerdude:
Hide music playback, soundcloud, youtube, and all the pdfs

PR #424, opened 2020-07-28
Adds network monitor and changes buttons available based on internet

finben:
Tested on real device and works nicely. 

https://learnappmaking.com/nwpathmonitor-internet-connectivity/

Addresses #423


Hymns/AppDelegate.swift
```diff
@@ -9,9 +10,12 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     static let uiTestingFlag = "-UITests"
     #endif
 
+    static let networkMonitor = NWPathMonitor()
+
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
         // Override point for customization after application launch.
         FirebaseApp.configure()
+        AppDelegate.networkMonitor.start(queue: DispatchQueue.global(qos: .background))
```
skywalkerdude, line 18: Interesting. I didn't know you could monitor the network like this.

Could you change SystemUtil#isNetworkAvailable to use this AppDelete thing instead? Then use system util throughout the code instead of this static AppDelegate.

That way we can mock and unit test (See how I do it with HymnsRepository/Spec)

HymnsTests/Display/DisplayHymnViewModelSpec.swift
```diff
@@ -122,11 +127,7 @@ class DisplayHymnViewModelSpec: QuickSpec {
                         }
                         context("is not favorited") {
                             beforeEach {
-                                given(favoriteStore.isFavorite(hymnIdentifier: classic1151)) ~> { _ in
-                                    Just(false).mapError({ _ -> ErrorType in
-                                        .data(description: "This will never get called")
-                                    }).eraseToAnyPublisher()
-                                }
```
skywalkerdude, line 129: @finben your test is breaking because you deleted this

PR #425, opened 2020-07-29
Comment button

finben:


finben:
This was fun to build out, but on second thought I don't really like it. I feel like it is just borrowing too much blatantly from hymnal.net. I think we can just go forward without a comment button. It's a cool, somewhat novel idea. But after thinking more about it, it's probably really not necessary to have comments on hymns. I am fine dropping the comment feature idea entirely.

PR #430, opened 2020-08-06
Add new tag looks/behavior

skywalkerdude:


Hymns/Tags/TagSheetView.swift
```diff
@@ -42,21 +42,23 @@ struct TagSheetView: View {
...
-                                    Image(systemName: "checkmark.circle.fill")
+                                    Image(systemName: "xmark.circle.fill")
                                 }
                                 .tagPill(backgroundColor: tag.color.background, foregroundColor: tag.color.foreground)
                             }).padding(2)
                         }
                     }
+                    if self.viewModel.tagsForHymn.isEmpty && !self.viewModel.otherTags.isEmpty {
+                        Text("Tags").font(.headline).fontWeight(.bold)
+                        Text("\n ")
```
finben, line 53: Imo it looks much better if tags is still there empty when there are no tags instead of disappearing. Since no matter what the user has come to this page to do something with tags. This gives a more consistent feel to the page since "Tags" is always at the top of the sheet instead of appearing and disappearing when tags are added or removed.

Hymns/Tags/TagSheetView.swift
```diff
@@ -66,8 +68,8 @@ struct TagSheetView: View {
                     }, label: {
                         HStack {
                             Text("Create a new tag").font(.headline).fontWeight(.bold)
-                            Image(systemName: self.viewModel.showNewTagCreation ? "chevron.down" : "chevron.up")
-                        }.foregroundColor(.primary).padding(.vertical)
+                            Image(systemName: self.viewModel.showNewTagCreation ? "chevron.up" : "square.and.pencil")
+                        }.foregroundColor(self.viewModel.showNewTagCreation ? .primary : .accentColor).padding(.vertical)
```
finben, line 72: In the case where the user is going to need to click the button "create a new tag" to bring in the drop down. I think it is good to make it more clear we have a button here with functionality. The accent color in this case makes this much more apparent. I did change to square.and.pencil as well which I think is a much more clear button. However, you all are free to disagree and override this change.

Hymns/Tags/TagSheetView.swift
```diff
@@ -88,7 +90,7 @@ struct TagSheetView: View {
                         }.padding(.top)
                     }
                     Spacer()
-                }.offset(y: self.kGuardian.slide).animation(.easeInOut(duration: 1.0))
+                }.offset(y: self.kGuardian.slide)//.animation(.easeInOut(duration: 1.0))
```
finben, line 93: This animation was causing all kinds of problems. It was applying to everything on the view. So for instance every time you pull up tags. The Empty Tag Illustration photo was easing out even though it should never appear in the first place. Did you mean to apply this animation to something in particular?

Hymns/Tags/TagSheetViewModel.swift
```diff
@@ -63,7 +63,7 @@ class TagSheetViewModel: ObservableObject {
                 self.otherTags = allTags.filter({ tag -> Bool in
                     return !tagsForHymn.contains(tag)
                 })
-                if self.tagsForHymn.isEmpty || self.otherTags.isEmpty {
+                if self.tagsForHymn.isEmpty || self.otherTags.count < 5 {
```
finben, line 66: I think it makes sense that once a user has a whole bunch of tags we don't show them the create a new tag by default but instead just let them pick from what they have? Seeing as once I have a bunch of tags on something like Evernote for example I become more prone to just select from what I have rather than keep creating new tags. I may be wrong here though and 5 is admittedly an arbitrary number. Just bringing this up for discussion.

finben:
Lol, I basically added a commit and then did a review on my own work to explain the changes.... All your code is fine. I approve. Maybe, I should have just made another branch for my changes. idk. Not too many though.

PR #441, opened 2020-08-11
Do not show minimize caret on the search screen and only show the tool tip if they haven't seen it before AND the caret is up

finben:


Hymns/Display/SoundCloudView.swift
```diff
@@ -41,7 +43,11 @@ struct SoundCloudView: View {
                                                     value.indicatorAnchor = anchor
                     })
                 }
-                SoundCloudWebView(url: self.url)
+                SoundCloudWebView(url: self.url).onReceive(self.viewModel.activeTimer) { _ in
+                    if AVAudioSession.sharedInstance().secondaryAudioShouldBeSilencedHint {
+                        self.nowShowToolTip = true
+                    }
+                }
```
skywalkerdude, line 50: hey this is a good idea! To use a timer to approach this. I was thinking to just inspect the url, but I think this might be better.

I think we should stick with the `showSoundCloudMinimizeTooltip` and do all this other stuff in the view model.

So in the view model, basically do:

1) If `userDefaultsManager.showSoundCloudMinimizeTooltip` is true, start timer
2) every 1 second, poll for `AVAudioSession.sharedInstance().secondaryAudioShouldBeSilencedHint`
3) if true, animate `showSoundCloudMinimizeTooltip` to true
4) on dismiss: cancel timer and set `userDefaultsManager.showSoundCloudMinimizeTooltip` to false to save it

another thing that would be nice is to invert and rename `userDefaultsManager.showSoundCloudMinimizeTooltip` to `userDefaultsManager.hasSeenSoundCloudMinimizeTooltip`, since that is semantically what we are tracking in UserDefaults.

finben: Okay, I did most of this stuff. The problem I had is I couldn't find out how to receive from the timer in the viewModel for putting step 2 into the viewModel. Everything else though looks good and is working.

Issue #219, opened 2020-05-30
Have music player use another languages mp3 path if their own path is nil


finben:
Currently, a hymn in certain languages like French often do not have a musicMp3 path so their music player is not shown up since the path is nil. Ideally, if that same song has another language like English with a valid mp3 path we could just use that path for the music player since the music is language agnostic.

PR #460, opened 2020-08-12
Allows music to play for other languages by keeping the English URL

finben:
Addresses #219 

You may be able to find a more clever way to do this. But this works

### DIFF
--- a/Hymns/Display/DisplayHymnBottomBarViewModel.swift
+++ b/Hymns/Display/DisplayHymnBottomBarViewModel.swift
@@ -19,6 +19,7 @@ class DisplayHymnBottomBarViewModel: ObservableObject {
     private let mainQueue: DispatchQueue
     private let repository: HymnsRepository
     private let systemUtil: SystemUtil
+    private var workingMP3: URL?
 
     private var disposables = Set<AnyCancellable>()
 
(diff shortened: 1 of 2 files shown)


Hymns/Display/DisplayHymnBottomBarViewModel.swift
```diff
@@ -54,18 +56,22 @@ class DisplayHymnBottomBarViewModel: ObservableObject {
...
+                        self.storedMP3Path = mp3Url
+                    }
+
+                    let languages = self.convertToSongResults(hymn.languages)
+                    if !languages.isEmpty {
+                        buttons.append(.languages(languages))
+                    }
+
+                    if let mp3URL = self.storedMP3Path, self.systemUtil.isNetworkAvailable() {
+                        buttons.append(.musicPlayback(AudioPlayerViewModel(url: mp3URL)))
```
skywalkerdude, line 74: I think you need to at least first check if the current song has a mp3, and if it does, use that. There are certain songs where the Chinese version has a different tune than the English version. So at most we should use this as a fallback

Hymns/Display/DisplayHymnBottomBarViewModel.swift
```diff
@@ -27,14 +28,15 @@ class DisplayHymnBottomBarViewModel: ObservableObject {
          hymnsRepository repository: HymnsRepository = Resolver.resolve(),
          mainQueue: DispatchQueue = Resolver.resolve(name: "main"),
          backgroundQueue: DispatchQueue = Resolver.resolve(name: "background"),
-         systemUtil: SystemUtil = Resolver.resolve()) {
+         systemUtil: SystemUtil = Resolver.resolve(), workingURL: URL? = nil) {
```
skywalkerdude, line 31: I generally don't like passing parameters around because if added 1-2 parameters for every new feature, the app would get unwieldy very quickly. Can you think of a better way? Maybe we can use the repository and fetch the English song version asynchronously if the current mp3 doesn't exist.

PR #463, opened 2020-08-14
on IPad users play songs right from search url. So carat should always be up

finben:


Hymns/Display/SoundCloudViewModel.swift
```diff
@@ -17,19 +17,28 @@ class SoundCloudViewModel: ObservableObject {
...
     }
 
     init(url: URL) {
         self.url = url
+        if UIDevice.current.userInterfaceIdiom == .pad {
+            self.showMinimizeCaret = true
+            if !self.hasSeenSoundCloudMinimizeTooltip {
+                self.showMinimizeToolTip = true
+            }
+        }
```
skywalkerdude, line 41: Can you put this in the urlObserver block instead of here in the init?

finben: No, I tried that at first. The problem is it doesn't work once you minimize and come back when I put it there.

skywalkerdude: Generally speaking, when something doesn't work a particular way, it's more important to know why it doesn't work that way than to just find another way where it happens to "work."

I tested this also on a phone simulator, and there's also the same problem -- when you minimize and open again, the caret is gone. So I'm not sure that the problem is that it's using an iPad. Are you testing on a device or simulator? It might be an issue with the simulator, but I'm not sure because I don't have a real ipad to test on 😞

skywalkerdude:
What if we went back to the timer approach? Whenever we get `silenceSecondaryAudioHintNotification`, then we show the caret. The only downside would be, if they had spotify or something playing already as they use our app, it would think that the soundcloud was playing when it wasn't playing. But do you think it's a good trade-off?

(Or even better, can you think of a way to narrow down that it is FOR SURE our soundcloud web view that is playing?)

Hymns/Display/SoundCloudViewModel.swift
finben: @skywalkerdude So there are really two problems here. The first is that the iPad was never using if path != search because on iPad the users do not navigate into the songs to play them. They simply play them right from the search. 

The second problem is worse and I thought was only on iPad but you are right it is on both! We need to get this resolved ASAP. Honestly, maybe we should even rethink ever not showing the carat? Especially, since iPad is going to always have to show the carat anyways. I don't know how much value we are getting out of hiding the carat anyways. May just be more confusing to users?

PR #464, opened 2020-08-14
It needs an if statement to reevaluate logic. Otherwise the view is lazy

finben:
Just needed an if statement to reevaluate logic. Otherwise the view is lazy

This took forever to debug. lol. This is a fix though. For the error you showed me the other day when the bottom bar vanishes after 1. going to another tune or language. 2. Coming back through back navigation 3. going to soundcloud 4. closing soundcloud

### DIFF
--- a/Hymns/Display/DisplayHymnView.swift
+++ b/Hymns/Display/DisplayHymnView.swift
@@ -27,8 +27,10 @@ struct DisplayHymnView: View {
                     } else {
                         viewModel.currentTab.content
                     }
-                    viewModel.bottomBar.map { viewModel in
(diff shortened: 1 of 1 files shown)


finben:
@skywalkerdude Not totally, whenever you leave SoundCloud to go back to DisplayHymn in these edge cases. The view is not being initialized again since it was only hidden. So I think the bug was related to that. Without the if statement the viewModel.bottombar.map function doesn't get reevaluated and repopulate the bottomBar. Kinda wierd, I know SwiftUI views try to be as lazy as possible for optimization so it may just be this sort of thing doesn't get reevaluated without the if block. 

lol

PR #468, opened 2020-08-14
Undo fix that didn't work and replace with fix that works 

finben:


finben:
Lmk if you can find something better. But this fixes it now for me. :0

Issue #461, opened 2020-08-12
Create settings option to clear someone's history/recent songs


skywalkerdude:
Create settings option to clear someone's history/recent songs and show a toast if it succeeds or fails

PR #469, opened 2020-08-14
Add setting to clear history

skywalkerdude:
fix https://github.com/HymnalDreamTeam/Hymns-iOS/issues/461


skywalkerdude:
@finben can you translate the German and Spanish strings? (should be 4 each)

finben:
It would be nice if clearing userDefaults wouldn't kick them out of the app. Can we just set them all to their default values when they do that?

finben:
Also, there should be some sort of warning before they delete them in case someone misclicks

PR #472, opened 2020-08-19
Soundcloud player enhancements

skywalkerdude:
1. Use a timer to determine when to show the minimize caret and tool tip
2. Listen to the title from the web view and add a marquee view to display it


skywalkerdude:
@finben check it out. this is pretty neat

PR #499, opened 2020-09-15
Add option to change the speed of songs

finben:


finben:
Simulator Screen Shot - iPhone 11 Pro - 2020-09-15 at 15 27 19

finben:
Probably need Emily to go over the design lol. It works nice though. It sounds a bit choppy when they slow down to song. But it is nice for fine tuning especially around the 1x mark like to go a bit faster.

skywalkerdude:
I see. Nice! @Emyeh What do you think about the screenshot?

Emyeh:
Nice! Let’s put it on the left side so that the main player stays centered. Add 8dp padding between the main player and speed controls. It looks too tight right now.

Now that are more controls I think we’ll need a thin divider that sits on top of the bottom toolbar. Then it’ll easier to distinguish between that and the control above.

skywalkerdude (approved):

PR #513, opened 2020-09-27
Disable swiping between songs if the AudioPlayer/SoundCloud modal is up

skywalkerdude:
Both the Audio Player and the SoundCloud modal has a horizontal scrubber that will get messed up if pager swipe is enabled


skywalkerdude:
@finben could you test this as well on your device? Make sure that minimizing the soundcloud player to the mini player still works as expected (since we can't test that on simulator)

PR #521, opened 2020-11-02
Add translation for two new strings

skywalkerdude:
@finben pls help translate


finben (commented):
I don't see the German file. But go ahead and just add this to the German string file

In die Zwischenablage kopierter Vers -> verse copied to clipboard
Geschwindigkeit -> speed

PR #536, opened 2021-01-09
Fix bug where when we try to go to a song that was greater than that hymn type's max number, it would go to the max number instead.

skywalkerdude:


Hymns/Display/DisplayHymnContainerViewModel.swift
```diff
@@ -31,7 +31,7 @@ class DisplayHymnContainerViewModel: ObservableObject {
         let hymnNumber = identifier.hymnNumber
         let queryParams = identifier.queryParams
 
-        if hymnType.maxNumber > 0 && hymnNumber.isPositiveInteger, let hymnNumberInt = hymnNumber.toInteger {
+        if hymnType.maxNumber > 0 && hymnNumber.isPositiveInteger, let hymnNumberInt = hymnNumber.toInteger, hymnNumberInt <= hymnType.maxNumber {
```
skywalkerdude, line 34: @Jason right here

PR #541, opened 2021-05-11
remove the unnecessary linker flag related to FIRanalyticsConnector an…

finben:
…d let pods update instead of locking down to 13.2


skywalkerdude (dismissed):

Hymns/Browse/AllSongsViewModel.swift
```diff
@@ -12,6 +12,6 @@ class AllSongsViewModel: ObservableObject {
 
 extension Resolver {
     public static func registerAllSongsViewModel() {
-        register {AllSongsViewModel()}.scope(graph)
+        register {AllSongsViewModel()}.scope(.graph)
```
finben, line 15: fix a deprication warning

PR #546, opened 2021-09-23
fix the nav pop on recent by using btn in lieu of nav link directly

finben:
This fixes the recents list problem. You probably can find a more elegant solution than me but at least this gets us working again and can point you in the right direction.

So basically what I think is happening is the following. In HomeView we have navigationLinks within a list. But that list is driven by viewModel.songResults. Well, when we click a recent song to go and see it then that calls an update to the list to re-add it again to the recent song list. But because of the way navigationLinks work in Lists or ForEach this change to the data source array songResults causes the view to be redrawn and gives us the popping issue. 

This workaround is to use a button in the List instead of Navigation Links directly. The button simply changes the route and then triggers navigation. This solves the problem because buttons do not seem to have the same issue when populated by a list or foreach which has a changing array of items.

### DIFF
--- a/Hymns/Display/DisplayHymnBottomBar.swift
+++ b/Hymns/Display/DisplayHymnBottomBar.swift
@@ -6,6 +6,7 @@ struct DisplayHymnBottomBar: View {
     @Binding var dialogModel: DialogViewModel<AnyView>?
     @State private var actionSheet: ActionSheetItem?
     @State private var sheet: DisplayHymnSheet?
+    @State private var fullSheet: DisplayFullSheet?
 
     // Navigating out of an action sheet requires another state variable
     // https://stackoverflow.com/questions/59454407/how-to-navigate-out-of-a-actionsheet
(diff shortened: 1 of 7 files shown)


finben:
For the ipad fix this does move us away from the bottom tabs. To me that is fine and actually gives a better and more consistent iPad experience and seems to be the way Apple is suggesting by looking at their own iPad apps. But it is worth noting. The second column with the navigation options can either be dismissed via tapping or swiping. Here are some screenshots.
Simulator Screen Shot - iPad Pro (11-inch) (3rd generation) - 2021-09-23 at 17 19 17
Simulator Screen Shot - iPad Pro (11-inch) (3rd generation) - 2021-09-23 at 17 18 33
Simulator Screen Shot - iPad Pro (11-inch) (3rd generation) - 2021-09-23 at 17 19 10

finben:
@skywalkerdude I think I stumbled on a solution for tag issue by just using something called .fullScreenCover.  This seems to deal with the issue and I don't see any more bugs with tag sheet now.

PR #551, opened 2021-09-24
Add an id to the List of views in HomeView so it wouldn't automatical…

skywalkerdude:
…ly pop back out whenever it changes (https://stackoverflow.com/questions/58376626/swiftui-navigation-automatically-closing-pop-realm)

### DIFF
--- a/Hymns/Home/HomeView.swift
+++ b/Hymns/Home/HomeView.swift
@@ -64,7 +64,7 @@ struct HomeView: View {
                     }.padding(.horizontal)
                     Spacer()
                 } else {
-                    List(viewModel.songResults) { songResult in
+                    List(viewModel.songResults, id: \.self) { songResult in
                         NavigationLink(destination: songResult.destinationView) {
                             SongResultView(viewModel: songResult)
                         }.onAppear {


finben:
It's wierd. I tried this and don't remember it working. But for some reason this works now. Maybe I did it a bit different. But it looks like this solves the pop problem?

finben:
Okay, after spending a couple hours on this, I'm throwing in the towel. I tried a bunch of things to little avail. I tried another hash function, converting over to a struct, using @ObservedObject and passing to a middle view. Basically nothing works.

 I think what happens is when we click a recent song there is a moment when the array has a duplicate of that song as it is readded to the top of the recents list. Because ForEach and list always require that no item has an exact duplicate to guarantee unique items this freaks it out and pops the navigation... But I don't really know.

finben:
I made a change on my original branch "recent_fix" that you should look at. I basically made the label of the button a navigation view. It now looks 100% like a navigation link and even seems to have sped up the rendering significantly. With this in I don't even see the benefit of continuing to look for another solution for the nav link.

